PR #5, opened 2018-01-18
Drop parse error for missing </caption> end tag

sideshowbarker:
See https://html.spec.whatwg.org/multipage/parsing.html#parsing-main-incaption:stack-of-open-elements-2
The parse error for this case was dropped from the spec in
https://github.com/whatwg/html/commit/759ee62

Relates to https://github.com/validator/validator/issues/593
and https://bugzilla.mozilla.org/show_bug.cgi?id=1429415


sideshowbarker:
Fixed in  https://github.com/validator/htmlparser/commit/7583e1ced299e7d84f49de34a5d9744b6760e5d6

Issue #17, opened 2020-04-24
Merge commits from validator-nu branch to master (then delete validator-nu branch)


sideshowbarker:
I’d like to merge everything from the validator-nu branch to master, then delete the branch.

And then as far as the HTML checker, the parser integration for it could be built from the master branch here rather from from the validator-nu branch it has been getting built from.

So, looking at https://github.com/validator/htmlparser/compare/validator-nu, there are 19 commits in the validator-nu branch which aren’t in master. None have any merge conflicts with master.

(The validator-nu branch did previously have 20 commits; the additional commit was 2bbdc4d, which removed all the code for the XOM API — but that commit’s no longer in the branch.)

Among those 19 commits, 15 are changes to the error-reporting behavior of the parser.

And among those 15, the largest set are changes necessary to make the error-reporting behavior conform to current requirements in the HTML spec; those 7 commits are the following:

e03e56e Require UTF-8
9ce4bd4 Conform ampersand-error reporting to HTML spec
29a2645 Make consecutive hyphens in comments a non-error
f738d97 Drop parse error for missing end tag
1cd256d Emit error (not warning) for HTML4/XHTML1 doctype
da6c1ec Report error always for Transitional doctype
1ae6c63 Support “generate all implied end tags thoroughly”

The second-largest set of those error-reporting-behavior commits are 5 bug fixes:

003ad33 Fix typo in error message about over-deep tree
ef15099 Fix "non-space characters insided a table" typo
de301bc Fix grammar problem in HTML parser error message
f4f266c Correct error for EOF in “in template” state
7528857 Report 1024 as byte limit for meta charset sniff

So I think all 12 of those commits listed above could be considered necessary changes.

The remaining 3 changes to the error-reporting behavior, although not strictly necessary, are nonetheless pretty clearly improvements:

143a10b Improve message: bad start tag in noscript in head
5a209fd Remove warning about comments before doctype
5c8fe7a Stop reporting HTML4-specific parse errors

Then, along with the 15 changes above to the error-reporting behavior, there’s 1 change to the core code that, from discussion in #10, it seems we already clearly know we want in master too:

1357528 Ensure every Locator is also a Locator2

And then the remaining 2 changes to the code are minor ones that we don’t necessarily need but that, since the work’s already been done, I guess we should go ahead and merge:

1c66c6f Ensure Java8-runnable code even if Java9-compiled
895f606 Add suppress "unchecked" for a TreeBuilder method

[...]

PR #22, opened 2020-07-04
Deprecate HTML4-error-reporting methods/classes and update pom.xml for the 1.5 release

sideshowbarker:
This change deprecates all public methods related to reporting of HTML4-specific parse errors — as well as deprecating the entire `nu.validator.htmlparser.common.DoctypeExpectation` class, which is only used when reporting HTML4-specific parse errors.

Compare to https://github.com/validator/htmlparser/commit/5c8fe7a

This change also updates the `pom.xml` file in preparation for the 1.5 release.


pom.xml
```diff
@@ -215,7 +215,7 @@
     <dependency>
       <groupId>xom</groupId>
       <artifactId>xom</artifactId>
-      <version>1.1</version>
+      <version>1.3.5</version>
```
carlosame, line 209: Should be `<version>[1.3.5,)</version>`

Otherwise, you are forcing downstream projects to use a specific version of XOM (or to configure POM exclusions).

sideshowbarker: > Should be `<version>[1.3.5,)</version>`

Thanks — changed it to that

pom.xml
```diff
@@ -158,6 +149,7 @@
             <Bundle-SymbolicName>nu.validator.htmlparser</Bundle-SymbolicName>
             <Bundle-Version>${project.version}</Bundle-Version>
             <Bundle-RequiredExecutionEnvironment>J2SE-1.5</Bundle-RequiredExecutionEnvironment>
+            <Automatic-Module-Name>nu.validator.htmlparser</Automatic-Module-Name>
```
anthonyvdotbe, line 152: As explained in https://github.com/validator/htmlparser/issues/17#issuecomment-653795447, I believe this should be reverted.

pom.xml
```diff
@@ -214,7 +206,7 @@
     <dependency>
       <groupId>xom</groupId>
       <artifactId>xom</artifactId>
-      <version>1.1</version>
+      <version>[1.3.5,)</version>
```
anthonyvdotbe, line 209: As explained in https://github.com/validator/htmlparser/issues/17#issuecomment-653795447, I believe this should be reverted.

pom.xml
anthonyvdotbe: I'm fine with this, no need to revert (see https://github.com/validator/htmlparser/issues/26)

pom.xml
anthonyvdotbe: I'm fine with the dependency upgrade (see https://github.com/validator/htmlparser/issues/26)

However, I hold that the use of version ranges should be reverted, as it's not a common Maven practice. See [here](https://github.com/hibernate/hibernate-validator/blob/master/pom.xml#L114) for a typical Maven example. As already mentioned [here](https://github.com/validator/htmlparser/issues/17#issuecomment-653795447), I believe fixed versions should be used, among other things "for the sake of build reproducibility and stability".

anthonyvdotbe: Ok, "common" was a bad choice of words. Let me state it differently: I very rarely, if ever, come across Maven POMs that use version ranges for dependencies.

From a quick look at some of the most popular Java libraries/frameworks on GitHub that use Maven, I don't find any that use version ranges for dependencies: [Maven](https://github.com/apache/maven/blob/master/pom.xml) itself, [Guava](https://github.com/google/guava/blob/master/pom.xml), [Byte Buddy](https://github.com/raphw/byte-buddy/blob/master/pom.xml), [Spring Data JPA](https://github.com/spring-projects/spring-data-jpa/blob/master/pom.xml), [Hibernate Validator](https://github.com/hibernate/hibernate-validator/blob/master/pom.xml), [commons-lang](https://github.com/apache/commons-lang/blob/master/pom.xml), [Quarkus](https://github.com/quarkusio/quarkus/blob/master/build-parent/pom.xml), [Helidon](https://github.com/oracle/helidon/blob/master/pom.xml), ... (same holds for a few Gradle projects I looked at, such as [Spring Framework](https://github.com/spring-projects/spring-framework/blob/master/build.gradle))

Guava has nearly [197000 dependents](https://github.com/google/guava/network/dependents?package_id=UGFja2FnZS0xNzk4OTA4NTM%3D). Now I don't know how many of those use Maven, but if the lack of version ranges would really pose a problem for downstream consumers, I'm pretty sure Guava would be using them by now.

By using fixed versions, you always know for each commit which version was used for each dependency: so if it built 2 years ago, it'll still build now. And if 2 developers clone the repo and build any given commit, they're always going to have equivalent results w.r.t. dependencies.

And w.r.t. downstream users: say 2 years ago `LibA 2.1` was published, with a dependency on `LibB` declared as `[1.7.1,)`.
Now if today I declare `LibA 2.1` in my POM, and the build fails because `LibB` is at version `6.5.8`, which is incompatible with `LibA 2.1`, then what do I do? At this point, all I know is: "at least some version >= 1.7.1 that was available when LibA was published, is compatible". And since Maven can resolve very surprising versions, such as alpha/beta releases (there's even an open [bug](https://issues.apache.org/jira/browse/MNG-3092) about Maven resolving SNAPSHOT versions for non-SNAPSHOT bounds), it's even possible that `LibA 2.1` is not compatible with *any* released version of LibB.
On the other hand, if LibA would've used fixed versions, things would just work.

As for detecting compatibility issues with new versions of dependencies: I agree that it's better to detect this sooner rather than later, but I believe there are better ways to do so.
If we consider a downstream user again: if `htmlparser 1.5` declares `XOM [1.3.5,)`, then 6 months from now, some downstream user might file a bug saying `htmlparser 1.5` is not compatible with `XOM 1.3.6`. And then the burden would be on `htmlparser` to fix the issue, because it basically declared "any version >= 1.3.5 works". On the other hand, by declaring `XOM 1.3.5`, a downstream user might still file a bug, but it's clear that `htmlparser` is not at fault, and never claimed it was compatible with any version > 1.3.5.

carlosame: > if the lack of version ranges would really pose a problem for downstream consumers, I'm pretty sure Guava would be using them by now

I was told otherwise by a former Maven developer, although perhaps there was a misunderstanding. And very large projects are much less likely to use dependency ranges.

Maven dependency version resolution is supposed to be somewhat flexible, and I'm not sure under which circumstances the problem would happen, but I was told that the safe approach (in a complex environment where fixed versions of the same dependency were specified by multiple libraries) was to configure exclusions. That may affect very big projects only though, and I haven't really seen a detailed account of it.

> all I know is: "at least some version >= 1.7.1 that was available when LibA was published, is compatible"

A dependency range does not give less information than a fixed one. If you know that the minimum version is `1.7.1`, that means you can build with exactly that one. Otherwise, the range is wrong.

> since Maven can resolve very surprising versions, such as alpha/beta releases

And such a resolution problem would affect the upstream developer, not downstream. If a beta is not compatible, that's a bug that you may have to file.

> it's better to detect this sooner rather than later, but I believe there are better ways to do so.

In css4j I have detected some important upstream issues by this method. BTW this requires a significant amount of tests that are able to detect such problems (about 45% of the css4j code is test-related).

> then 6 months from now, some downstream user might file a bug saying `htmlparser 1.5` is not compatible with `XOM 1.3.6`

That's the other part of the issue: in my opinion, users have the right expectation that a supported project is going to run with reasonably recent dependencies, and if that is not the case they sometimes may file a bug with the dependency, sometimes with you. I have had a few of those, and have to deal with that as a maintainer.

This project can choose to depend on a fixed version (and some people are going to understand that the library is endorsing that version, or can only work with it), or specify a range (and other people are going to have ridiculous expectations about what that means, as you mention).

The range could be more conservative than what I suggested: for example `[1.3.5,2)` could be safer for XOM. I specified a range for two of the dependencies (XOM and ICU4J, no plan to do the same with others), and these are unlikely to cause problems in the future. ICU4J numbering in particular is growing fast (currently at `67.1` in Maven Central).

anthonyvdotbe: > A dependency range does not give less information than a fixed one. If you know that the minimum version is `1.7.1`, that means you can build with exactly that one. Otherwise, the range is wrong.

Yes, if you first build with a fixed version of `1.7.1` to verify a successful build, and immediately afterwards commit with a version range like `[1.7.1,)`, then you know that the left bound works *for that specific commit*. But any commit after that may quietly break compatibility with `1.7.1`, since Maven might be resolving a version other than `1.7.1`. So I believe my point still holds.

> And such a resolution problem would affect the upstream developer, not downstream. If a beta is not compatible, that's a bug that you may have to file.

Maybe my terminology is off, but by my definitions, I'm a downstream user of LibA, which is in turn a downstream user of LibB. And the example shows that I can't easily add LibA as a dependency, due to its use of version ranges instead of fixed versions.

As a final argument: Karl Heinz Marbaise [says](https://stackoverflow.com/questions/44521542/why-doesnt-gradle-or-maven-have-a-dependency-version-lock-file#comment76075193_44525378) this on version ranges (emphasis is mine): "It is not a question of could define versions ranges in dependencies. It can be done...But as I explained **it is not recommended...I would never do so**...and for lazy devops people there are tools which can handle this..."

carlosame: > But any commit after that may quietly break compatibility with 1.7.1

Developers use fixed dependency versions in their IDEs (I certainly do), and chose carefully to which APIs they conform (or at least they should). I only use Maven to package and deploy (to Github Packages in my case), and my understanding is that a lot of people (including the maintainers of this project) do similarly.

I mean, I think of the Maven POM as something largely for "external consumption".

> the example shows that I can't easily add LibA as a dependency, due to its use of version ranges instead of fixed versions

The final application should use specific versions and, if they really do not know better, then use for LibB the lower range that was specified by LibA.

> Karl Heinz Marbaise says this on version ranges

He mentions _devops_ and deployment tools, so I presume that he is talking about the whole application. Yes, final downstream applications should always define specific versions of the dependencies, we all agree on this (I told you just that elsewhere). And this project is a library that tries to not get in the way of that process.

anthonyvdotbe: > He mentions _devops_ and deployment tools, so I presume that he is talking about the whole application. Yes, final downstream applications should always define specific versions of the dependencies, we all agree on this (I told you just that elsewhere). And this project is a library that tries to not get in the way of that process.

The OP asks:
> Are the packages that exist within the Maven Central Repository [...], allowed to have version ranges for *their* dependencies defined in their pom.xmls?

To which he first replies:
> It is allowed [...] But [...] this by definition makes your own build and builds of your users not reproducible [...]

In a followup comment, he reiterates his point:
> It is not a question of could define versions ranges in dependencies. It can be done...But as I explained **it is not recommended...I would never do so**...and for lazy devops people there are tools which can handle this... 

So this is in the context of Maven Central. And Maven Central mainly holds libraries, not "final downstream applications".

Additionally, I gave you a list of some of the most popular libraries/frameworks above already, none of which use version ranges. From what I understand, you feel that the given list wasn't representative somehow, so please do share a list of popular Java libaries that you do believe are representative and use version ranges.

carlosame: > > makes your own build and builds of your users not reproducible

As I explained, Maven builds are by default not reproducible, and those "builds of your users" _should always have fixed versions configured by those users_.

I insist that this is not something where I have a strong opinion (unlike the other details of the modularization), but I see that line of thought somewhat confusing (to not say "confused").

> > for lazy devops people there are tools which can handle this

AFAIK there is no such a thing as a deployment tool that 'handles dependencies' for an isolated library.

> list of popular Java libaries that you do believe are representative and use version ranges

I am not aware of popular libraries that use ranges, but I have seen ranges used for internal components in application servers, and 'popularity' is something that has to be put always in perspective.

I expect to see more in the future, as projects are being modularized, but in any case a modular JDK is going to be _de-facto_ restricted to XOM versions over `1.3.5` and ICU4J over `64.2`.

The main blocker here is that -as mentioned- having fixed versions in the POM apparently represents a problem for downstream users, which was a surprise to me when I was made aware of it (I omit the product name 'to protect the innocent'):

> Without upgrading to dom4j `2.1.3` `OUR_PRODUCT` wouldn't build since this version of css4j depends on dom4j `2.1.3` (in its pom) and this `OUR_PRODUCT`'s build was conflicting and telling us that we 2 different versions of Dom4J in our various pom.xml

And yes those people do specify fixed versions for every and each of their dependencies, and are highly proficient. AFAIR they use a commercial 'deployment tool' which could be part of the problem, I don't really know.

carlosame: I finally found out what was going on with the fixed dependencies; it appears that for downstream users, setting this:

`<version>1.3.5</version>`

is effectively the same as:

`<version>[1.3.5,)</version>`

which is not what I read in:

https://maven.apache.org/pom.html#dependency-version-requirement-specification

"`1.0`: Soft requirement for 1.0. Use 1.0 if no other version appears earlier in the dependency tree."

nor in:

https://maven.apache.org/guides/introduction/introduction-to-dependency-mechanism.html#transitive-dependencies

"Maven picks the "nearest definition". That is, it uses the version of the closest dependency to your project in the tree of dependencies. You can always guarantee a version by declaring it explicitly in your project's POM. Note that if two dependency versions are at the same depth in the dependency tree, the first declaration wins."

**Nope, the real behaviour is described here:**

https://docs.oracle.com/middleware/1212/core/MAVEN/maven_version.htm#CJHDEHAB

"`1.0` | It generally means 1.0 or a later version, if 1.0 is not available. Various Maven plug-ins may interpret this differently, so it is safer to use one of the other, more specific options."

That is, Oracle recommends using version ranges...

More from that page:

"When Maven encounters multiple matches for a version reference, it uses the highest matching version. Generally, version references should be only as specific as required so that Maven is free to choose a new version of dependencies where appropriate"

anthonyvdotbe: Marbaise says on the same page:
> If you define the versions in your pom file the dependency tree is always the same **which means you don't need to define all transitive dependencies and you saved a lot of work. Except if you use versions ranges in Maven**

I hope we can agree that (a) this is not the point of view you're defending here, and (b) given Marbaise's Maven expertise, it's much more likely that his view is in line with best practices.

> I am not aware of popular libraries that use ranges, but I have seen ranges used for internal components in application servers, and 'popularity' is something that has to be put always in perspective.

I rest my case. It is beyond me why we're still having this discussion, if it's impossible to readily give me a list of projects that use version ranges and are at least somewhat popular. And just to give some verifiable counterexamples w.r.t. application servers: from a quick look, neither [WildFly](https://github.com/wildfly/wildfly/blob/master/pom.xml) nor [Payara](https://github.com/payara/Payara/blob/master/pom.xml) use version ranges.

> And yes those people do specify fixed versions for every and each of their dependencies, and are highly proficient.

Which, as mentioned above, is not Marbaise's point of view. And "are highly proficient" is an unverifiable opinion, whereas it's trivial to verify that Marbaise is a Maven expert.

> AFAIR they use a commercial 'deployment tool' which could be part of the problem, I don't really know.

I'd say this speaks for itself.

With that being said, I'm out for now.

anthonyvdotbe: > With that being said, I'm out for now.

Well, that was before I read your [latest comment](https://github.com/validator/htmlparser/pull/22#discussion_r451986815). Honestly, I'm mindblown by your ability to select and interpret information in such a way that it matches your beliefs.

That page is about Maven *in the context of Oracle Fusion Middleware*. And yes, in that context, it's reasonable to declare a dependency as `[12.1.2,12.1.3)`, because you don't want to have to bump the dependency version whenever the middleware server is patched. This is about using *strict* version ranges for a *very specific use case* in a *tightly-controlled environment*.

Also note that this is about the use of version ranges in *middleware applications*, not libraries.

As for the first part of your comment: if you believe it presents a valid point, please describe a concrete example (like: "A has a dependency on B, and B has a dependency on C. If A's `dependencies` section declares ... and B's `dependencies` section declares ..., then changing the section(s) as follows ..., will give equivalent results when doing a build of A"), or better yet: set up a repo with 3 folders A/B/C, each containing a Maven project, so I can see for myself what you mean.

So let me try that again: with that being said, I'm out for now.

carlosame: Mr. @anthonyvdotbe is "out" (and hopefully won't come back) and I'm tired of this thread as well, but if anyone wants to know about how those build failures (in complex projects) happen, there is an informative blog post here:

https://ourcraft.wordpress.com/2016/08/22/how-to-read-maven-enforcer-plugins-requireupperbounddeps-rule-failure-report/

My downstream project's build error was of the kind described there, and version ranges are an easy solution to the problem.

anthonyvdotbe: To summarize the thread so far:

Given the following dependency graph:
````
root -> htmlparser X -> xom 1.3.5
     -> libB Y -> xom 1.3.6
````

Then when `root`'s POM declares:
1. no xom dependency and
   1. `htmlparser` is declared first or `xom` is excluded from `libB`: PASS with 1.3.5
   2. `libB` is declared first or `xom` is excluded from `htmlparser`: PASS with 1.3.6
2. xom 1.3.5: PASS with 1.3.5
3. xom 1.3.6: PASS with 1.3.6

However, when enabling [this rule](https://maven.apache.org/enforcer/enforcer-rules/requireUpperBoundDeps.html) from the `maven-enforcer-plugin`, case 1.1 and 2 would FAIL instead.

Now @carlosame holds that `htmlparser` should depend on `xom [1.3.5,)`, since it allows case 1.1 to PASS, even with the rule enabled. However, I hold that version ranges should be avoided, and that the proper solution is for `root` to modify its POM to match 1.2. or 3 instead.

PR #24, opened 2020-07-05
Small POM fixes for 1.5

carlosame:
Two small POM updates for 1.5:

1. Update OSGi metadata for the new 1.7 environment.
2. Update `scm` information which still pointed to the old site.


carlosame:
Also cherry-picked commit aacbf4ca to be able to build with Java 8.

PR #29, opened 2020-08-03
Import patches from the validator-nu branch that have landed in m-c as of 2020-08-03

hsivonen:
These are the changesets picked from the validator-nu branch that have the corresponding Gecko patches on Gecko autoland at this time. I suggest merging this when autoland merges to mozilla-central. (It's possible that there's a fancier way to do this with git, and I'm just bad at git.)


sideshowbarker:
> I suggest merging this when autoland merges to mozilla-central.

Sounds good — so for now I’ve moved this PR to “draft” status (which will prevent us from accidentally merging it in the meantime). As soon as autoland merges to m-c, we can flip this back to normal/non-draft status so it can be merged.

hsivonen:
And it turns out there's something in here that the translator does not like. I'll try to figure out what.

hsivonen:
> And it turns out there's something in here that the translator does not like. I'll try to figure out what.

Now resolved. I had forgotten the patch I wrote myself a month ago.

sideshowbarker:
> > And it turns out there's something in here that the translator does not like. I'll try to figure out what.
> 
> Now resolved. I had forgotten the patch I wrote myself a month ago.

I know the feeling 😄

PR #31, opened 2020-08-05
Fix TokenizerTester casing of PCDATA state name

sideshowbarker:
This change updates the `TokenizerTester` code to expect its input test data to have the string `Data state` to identify PCDATA tests — rather than the string `DATA state`.

The test data in the html5lib-tests suite uses `Data state`, so without this change, running TokenizerTester against html5lib-tests causes `TokenizerTester` to fail with a “Broken test data” harness failure.


hsivonen:
I pushed this via the command line as 11dbb0dcc14d780358846cd2cbbce0fba011eda2 and now I don't know how to make the GitHub PR metadata say merged instead of closed.

PR #32, opened 2020-08-05
Make TreeTester ignore tests for keygen element

sideshowbarker:
This change makes `TreeTester` ignore any tests which have the `keygen` element in their expected results.

Otherwise, without this change, we fail the tokenizer test at https://github.com/html5lib/html5lib-tests/blob/master/tree-construction/tests7.dat#L409 due to the fact with https://github.com/validator/htmlparser/commit/8f9f6bc we completely removed `keygen` parsing support from our code. (However, the `keygen` parsing requirements are still in the HTML spec, so the test case cannot be removed from html5lib-tests.)


hsivonen:
I pushed this via the command line as f3f92971004eaa1754121c36188984b85f174a3a and now I don't know how to make the GitHub PR metadata say merged instead of closed.

PR #34, opened 2020-08-06
Enable control for tokenizer buffer size

sideshowbarker:
This change adds an optional `bufferSize` parameter to the `tokenize()` `parse()` and `parseFragment()` methods of `sax.HtmlParser` instances. That `bufferSize` parameter controls the size of the buffer which gets fed to the tokenizer.

The control provided by that parameter allows the tokenizer buffer to be set, for example, to `1` — and that is particularly useful for emulating the behavior of the Firefox HTML parser, which feeds the tokenizer one single code unit at a time.

Otherwise, without this change, the tokenizer buffer size for `HtmlParser` instances is hardcoded to `2048`.

### DIFF
--- a/src/nu/validator/htmlparser/io/Driver.java
+++ b/src/nu/validator/htmlparser/io/Driver.java
@@ -173,6 +173,24 @@ public class Driver implements EncodingDeclarationHandler {
      *             if the stream threw
      */
     public void tokenize(InputSource is) throws SAXException, IOException {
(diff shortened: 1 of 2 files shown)


hsivonen:
I pushed this via the command line as 9ffe6b967b517baf42cdbec2daea03405ea4a3d1 and now I don't know how to make the GitHub PR metadata say merged instead of closed.

Issue #35, opened 2020-08-07
Investigate failing 30 tokenizer tests (including 10 tests related to handing of U+0000 characters)


sideshowbarker:
We’re failing 30 tests cases in https://github.com/html5lib/html5lib-tests/tree/master/tokenizer/ (see below). At least 10 of them are related to handling of U+0000 NUL characters.

I don’t understand why the Java parser is failing these but the Firefox parser isn’t.

```
--------------------------------
Failure
Raw NUL replacement
Input:
\u0000
Expected tokens:
[["Character","\\uFFFD"]]
Actual tokens:
[["Character","\\u0000"]]
--------------------------------
Failure
Raw NUL replacement
Input:
\u0000
Expected tokens:
[["Character","\\uFFFD"]]
Actual tokens:
[["Character","\\u0000"]]
--------------------------------
Failure
Raw NUL replacement
Input:
\u0000
Expected tokens:
[["Character","\\uFFFD"]]
Actual tokens:
[["Character","\\u0000"]]
--------------------------------
Failure
Raw NUL replacement
Input:
\u0000
Expected tokens:
[["Character","\\uFFFD"]]
Actual tokens:
[["Character","\\u0000"]]
--------------------------------
Failure
NUL in CDATA section
Input:
\u0000]]>
Expected tokens:
[["Character","\\u0000"]]
Actual tokens:
[["Character","\\u0000]]>"]]
--------------------------------
Failure
NUL in script HTML comment
Input:
<!--test\u0000--><!--test-\u0000--><!--test--\u0000-->
Expected tokens:
[["Character","<!--test\\uFFFD--><!--test-\\uFFFD--><!--test--\\uFFFD-->"]]
Actual tokens:
[["Character","<!--test\\u0000--><!--test-\\u0000--><!--test--\\u0000-->"]]
--------------------------------
Failure
NUL in script HTML comment - double escaped
Input:
<!--<script>\u0000--><!--<script>-\u0000--><!--<script>--\u0000-->
Expected tokens:
[["Character","<!--<script>\\uFFFD--><!--<script>-\\uFFFD--><!--<script>--\\uFFFD-->"]]
Actual tokens:
[["Character","<!--<script>\\u0000--><!--<script>-\\u0000--><!--<script>--\\u0000-->"]]
--------------------------------
Failure
lowercase endtags
Input:
</XMP>
Expected tokens:
[["EndTag","xmp"]]
Actual tokens:
[["Character","</XMP>"]]
--------------------------------
[...]

PR #36, opened 2020-08-10
Ensure testharness makes missing doctype name null

sideshowbarker:
This change ensures that the tokenizer sets the doctype name to null when the doctype name is missing in the input source.

Otherwise, without this change, the doctype name is set to the empty string — which doesn’t conform to the requirements in the HTML spec, and which causes us to fail 9 tests in the html5lib-tests suite.

This PR branch also includes a related change for the testharness.

Relates to https://github.com/validator/htmlparser/issues/35

--- 

The tests that are failing without this change are the following:

* https://github.com/html5lib/html5lib-tests/blob/master/tokenizer/test2.test#L3 (“DOCTYPE without name”)
* https://github.com/html5lib/html5lib-tests/blob/master/tokenizer/test3.test#L1956 (`<!DOCTYPE\\u0009`)
* https://github.com/html5lib/html5lib-tests/blob/master/tokenizer/test3.test#L1963 (`<!DOCTYPE\\u000A`)
* https://github.com/html5lib/html5lib-tests/blob/master/tokenizer/test3.test#L1970 (`<!DOCTYPE\\u000B`)
* https://github.com/html5lib/html5lib-tests/blob/master/tokenizer/test3.test#L1979 (`<!DOCTYPE\\u000C`)
* https://github.com/html5lib/html5lib-tests/blob/master/tokenizer/test3.test#L1986 (`<!DOCTYPE\\u000D`)
* https://github.com/html5lib/html5lib-tests/blob/master/tokenizer/test3.test#L2069 (`<!DOCTYPE  `)
* https://github.com/html5lib/html5lib-tests/blob/master/tokenizer/test3.test#L2153 (`<!DOCTYPE >`)
* https://github.com/html5lib/html5lib-tests/blob/master/tokenizer/test3.test#L5273 (`<!DOCTYPE>`)

### DIFF
--- a/src/nu/validator/htmlparser/impl/Tokenizer.java
+++ b/src/nu/validator/htmlparser/impl/Tokenizer.java
@@ -5942,7 +5942,7 @@ public class Tokenizer implements Locator, Locator2 {
         // Discard the characters "DOCTYPE" accumulated as a potential bogus
         // comment into strBuf.
         clearStrBufAfterUse();
-        doctypeName = "";
+        doctypeName = null;
         if (systemIdentifier != null) {
             Portability.releaseString(systemIdentifier);
             systemIdentifier = null;
@@ -6246,7 +6246,7 @@ public class Tokenizer implements Locator, Locator2 {
                          * Create a new DOCTYPE token. Set its force-quirks flag
                          * to on.
                          */
-                        doctypeName = "";
+                        doctypeName = null;
                         if (systemIdentifier != null) {
                             Portability.releaseString(systemIdentifier);
                             systemIdentifier = null;
--- a/test-src/nu/validator/htmlparser/test/JSONArrayTokenHandler.java
+++ b/test-src/nu/validator/htmlparser/test/JSONArrayTokenHandler.java
@@ -94,7 +94,7 @@ public class JSONArrayTokenHandler implements TokenHandler, ErrorHandler {
         flushCharacters();
         JSONArray token = new JSONArray();
         token.getValue().add(DOCTYPE);
-        token.getValue().add(new JSONString(name));
+        token.getValue().add(name == null ? JSONNull.NULL : new JSONString(name));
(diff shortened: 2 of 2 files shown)


hsivonen (approved):
Thanks. I believe there's another layer of bug / spec change in the tree builder on this topic.

sideshowbarker:
> I believe there's another layer of bug / spec change in the tree builder on this topic.

OK — I guess we should have (add) a test to expose it? This is another area of the code where it’s not clear to me how it aligns with the relevant spec algorithms. But I somewhat just trust that if we are passing the existing tests, that means the implementation is functionally the same as spec prose. But of course that can only be as reliable if we’re confidant we have coverage for the spec requirements in the test suite…

sideshowbarker:
I am assuming that the approval for the changes in this PR does not mean *“Let’s go ahead and merge this to master”*. Right? I suppose we should document the workflow somewhere, but I have thus far understand that in order to merge to master here, there are some other steps that need to take place on the Firefox side — for example, a patch submitted against a BMO bug, and a successful try run, etc. — and then if that all looks good, then you come back here and either merge the PR to master yourself, or comment here with an explicit OK.

hsivonen:
> I suppose we should document the workflow somewhere, but I have thus far understand that in order to merge to master here, there are some other steps that need to take place on the Firefox side — for example, a patch submitted against a BMO bug, and a successful try run, etc. — and then if that all looks good, then you come back here and either merge the PR to master yourself, or comment here with an explicit OK.

Yeah, there needs to be a [BMO bug number](https://bugzilla.mozilla.org/show_bug.cgi?id=1541822), a successful try run, and, typically, test cases that Firefox CI runs. Since autoland on the Firefox side has a certain risk of backouts, I wait until a changeset makes it to mozilla-central before merging the corresponding changeset here.

Thanks for the patch and patience.

PR #37, opened 2020-08-10
Make io.Driver expose public dontSwallowBom(), and use it in the testharness

sideshowbarker:
This change makes `nu.validator.htmlparser.io.Driver` instances have a `dontSwallowBom()` method exposed.

Otherwise, without this change, it’s not possible for users of `io.Driver` to preserve the BOM in input.

This PR branch also has a related change for the testharness. Without that change, the following html5lib-tests case fails:

* https://github.com/html5lib/html5lib-tests/blob/master/tokenizer/domjs.test#L197 (“leading U+FEFF must pass through”)

### DIFF
--- a/src/nu/validator/htmlparser/io/Driver.java
+++ b/src/nu/validator/htmlparser/io/Driver.java
@@ -59,7 +59,7 @@ public class Driver implements EncodingDeclarationHandler {
      */
     private RewindableInputStream rewindableInputStream;
 
-    private boolean swallowBom;
+    private boolean swallowBom = true;
 
     private Encoding characterEncoding;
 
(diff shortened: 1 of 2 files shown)


src/nu/validator/htmlparser/io/Driver.java
```diff
@@ -266,7 +265,7 @@ public void tokenize(InputSource is) throws SAXException, IOException {
         }
     }
 
-    void dontSwallowBom() {
+    public void dontSwallowBom() {
```
carlosame, line 268: `setDontSwallowBom()` looks a bit more java-ish.

hsivonen (approved):
I think this is good as-is. `setDontSwallowBom()` would imply also being able to change the state back.

hsivonen:
I pushed this via the command line as a0c0a75648a9e9aaca61be044043a567af4483ca and now I don't know how to make the GitHub PR metadata say merged instead of closed.

PR #38, opened 2020-08-10
Set default lastStartTag in TokenizerTester

sideshowbarker:
This change causes `TokenizerTester` to set a default `lastStartTag` value for any test which has a `initialStates` value but no `lastStartTag` value.

Otherwise, without this change, the tokenizer is never actually run for such missing-`lastStartTag` tests — which results in us failing 5 tests from the html5lib-tests suite.

---

The html5lib-tests cases which fail without this change are the following:

* https://github.com/html5lib/html5lib-tests/blob/master/tokenizer/domjs.test#L28 (“Raw NUL replacement”)
* https://github.com/html5lib/html5lib-tests/blob/master/tokenizer/domjs.test#L38 (“NUL in CDATA section”)
* https://github.com/html5lib/html5lib-tests/blob/master/tokenizer/domjs.test#L284 (“CDATA content”)
* https://github.com/html5lib/html5lib-tests/blob/master/tokenizer/domjs.test#L290 (“CDATA followed by HTML content“)
* https://github.com/html5lib/html5lib-tests/blob/master/tokenizer/domjs.test#L296 (“CDATA with extra bracket”)

### DIFF
--- a/test-src/nu/validator/htmlparser/test/TokenizerTester.java
+++ b/test-src/nu/validator/htmlparser/test/TokenizerTester.java
@@ -140,18 +140,23 @@ public class TokenizerTester {
         } else {
             for (JSONValue value : contentModelFlags.getValue()) {
                 if (PCDATA.equals(value)) {
+                    lastStartTag = lastStartTag == null ? "xmp" : lastStartTag;
                     runTestInner(inputString, expectedTokens, description,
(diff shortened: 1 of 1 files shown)


hsivonen:
I pushed this via the command line as 23a8b13439f65a4e0cb963075cef78661aed264e and now I don't know how to make the GitHub PR metadata say merged instead of closed.

PR #39, opened 2020-08-10
Make TokenizerTester handle double-escaped tests

sideshowbarker:
This change makes `TokenizerTester` correctly handle tests in html5lib-tests suite which have cases with so-called “double-escaped” `input` and `outputs values — for example, values that contain the literals `\\u0000` and `\\uFFFD` rather than `\u0000` and `\uFFFD`.


hsivonen:
I pushed this via the command line as 7f0ae26a2bae5c5e4376e6b09a4b7b51890b7df9 and now I don't know how to make the GitHub PR metadata say merged instead of closed.

PR #40, opened 2020-08-11
Conform tokenizer-only U+0000 NUL handling to spec

sideshowbarker:
This change brings the tokenizer’s handling of U+0000 NUL characters in the DATA state and the CDATA section state into conformance with the requirements in the HTML spec — for the case where only tokenization is being performed, without tree construction; that is, the case where the `tokenizer()` method is called, rather than `parse()` or `parseFragment()`.

Specifically, the tokenization steps defined in the spec require that when a U+0000 NUL is consumed in the DATA state or in the CDATA section state, the parser must then emit a U+0000 NUL. But when performing tree construction, the spec requires that when a U+0000 NUL is consumed, the parser must instead emit a U+FFFD REPLACEMENT CHARACTER.

Without this change, the parser always emits a U+FFFD REPLACEMENT CHARACTER — even when only tokenization is being performed. That causes us to fail a number of tests in html5lib-tests suite.

For more background on the relevant behavior, see the following:

* https://www.w3.org/Bugs/Public/show_bug.cgi?id=9659
* https://github.com/whatwg/html/commit/d98f83e
* https://github.com/validator/htmlparser/commit/9b9c263

Relates to https://github.com/validator/htmlparser/issues/35

### DIFF
--- a/src/nu/validator/htmlparser/common/TokenHandler.java
+++ b/src/nu/validator/htmlparser/common/TokenHandler.java
@@ -145,6 +145,17 @@ public interface TokenHandler {
     public void zeroOriginatingReplacementCharacter() throws SAXException;
     
     /**
+     * Emits:
+     *
+     *  * U+0000 if only tokenization is being performed
+     *  * U+FFFD if tree construction is being performed also
+     *
+     * @throws SAXException
+     *             if something went wrong
+     */
+    public void zeroOrReplacementCharacter() throws SAXException;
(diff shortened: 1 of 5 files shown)


hsivonen:
Landed as 9d72e928e1e683f7eac9946678ed1b4a3d94175a before realizing there was a PR.

Thanks, and sorry about the resulting bad metadata on this PR>

PR #41, opened 2020-08-11
testharness: Use the correct lastStartTag value when running tests with “Script data state” as the initial state

sideshowbarker:
This change ensures that `TokenizerTester` uses the correct, expected `lastStartTag` value from the test source when the `initialStates` value in the test source includes `Script data state` — or else if the test source specifies no `lastStartTag`, then `script` is used as the value.

Otherwise, without this change, `TokenizerTester` is hardcoded to always use `script` as the `lastStartTag` when running a test case with “Script data state” as the initial state. That discrepancy causes us to fail the following case from the html5lib-tests suite:

* https://github.com/html5lib/html5lib-tests/blob/master/tokenizer/domjs.test#L219-L223 (“lowercase endtags”)


hsivonen:
I pushed this via the command line as f23387a8e09711a2a458bc8f4ea2eaec5ddb8fd8 and now I don't know how to make the GitHub PR metadata say merged instead of closed.

Issue #33, opened 2020-08-05
Investigate failing math-element tree-construction tests


sideshowbarker:
We’re failing four test cases in https://github.com/html5lib/html5lib-tests/blob/master/tree-construction/math.dat (see below). I don’t understand why the Java parser is failing these but the Firefox parser isn’t.

These seem related to whatwg/html@8cf34ad *“Handle foster-parenting of and elements”*, which as far as I can see, we maybe never implemented support for in the Java sources — in which case, it would be expected to see the tests failing just as I’ve observed them to be.

But if so, it’s again a mystery to me why these tests would be passing in Firefox but not when run with the Java parser.

```
Data:
<math><tr><td><mo><tr>
Expected:
| <math math>
|   <math tr>
|     <math td>
|       <math mo>
Got:
| <math math>
|   <math tr>
|     <math td>
|       <math mo>
|   <math tr>
Expected errors:
Actual errors:
6: Start tag “math” seen in “table”.
22: End of file seen and there were open elements.
22: Unclosed element “tr”.
6: Unclosed element “math”.
-----------------------------------------
Data:
<math><thead><mo><tbody>
Expected:
| <math math>
|   <math thead>
|     <math mo>
Got:
| <math math>
|   <math thead>
|     <math mo>
|   <math tbody>
Expected errors:
Actual errors:
6: Start tag “math” seen in “table”.
24: End of file seen and there were open elements.
24: Unclosed element “tbody”.
6: Unclosed element “math”.
-----------------------------------------
Data:
<math><tfoot><mo><tbody>
Expected:
| <math math>
|   <math tfoot>
|     <math mo>
Got:
| <math math>
|   <math tfoot>
|     <math mo>
|   <math tbody>
Expected errors:
Actual errors:
6: Start tag “math” seen in “table”.
24: End of file seen and there were open elements.
24: Unclosed element “tbody”.
6: Unclosed element “math”.
-----------------------------------------
Data:
<math><tbody><mo><tfoot>
Expected:
| <math math>
|   <math tbody>
|     <math mo>
Got:
| <math math>
|   <math tbody>
|     <math mo>
|   <math tfoot>
Expected errors:
Actual errors:
6: Start tag “math” seen in “table”.
24: End of file seen and there were open elements.
24: Unclosed element “tfoot”.
6: Unclosed element “math”.
```

sideshowbarker:
> I don’t understand why the Java parser is failing these but the Firefox parser isn’t

[...]

PR #42, opened 2020-08-15
Ensure namespace-aware “clear the stack” handling

sideshowbarker:
This change ensures that for all cases with spec requirements in the form “clear the stack back to a *foo* context” — which involves checking for elements with particular names — we only look for elements in the HTML namespace, rather than additionally looking for elements which aren’t in the HTML namespace but that also have those particular names.

Otherwise, without this change, we aren’t in conformance with the spec requirements, and we fail several cases in the html5lib-tests suite.

Fixes https://github.com/validator/htmlparser/issues/33

---

Without this change, we’re failing the following html5lib-tests fragment-parsing cases:

* https://github.com/html5lib/html5lib-tests/blob/master/tree-construction/math.dat#L13 `<math><tr><td><mo><tr>` context: tr
* https://github.com/html5lib/html5lib-tests/blob/master/tree-construction/math.dat#L24 `<math><thead><mo><tbody>` context: thead
* https://github.com/html5lib/html5lib-tests/blob/master/tree-construction/math.dat#L34 `<math><tfoot><mo><tbody>` context: tfoot
* https://github.com/html5lib/html5lib-tests/blob/master/tree-construction/math.dat#L44 `<math><tbody><mo><tfoot>` context: tbody

…and also failing the corresponding tests in https://github.com/html5lib/html5lib-tests/blob/master/tree-construction/svg.dat

### DIFF
--- a/src/nu/validator/htmlparser/impl/TreeBuilder.java
+++ b/src/nu/validator/htmlparser/impl/TreeBuilder.java
@@ -3898,8 +3898,9 @@ public abstract class TreeBuilder<T> implements TokenHandler,
 
     private int findLastInTableScopeOrRootTemplateTbodyTheadTfoot() {
         for (int i = currentPtr; i > 0; i--) {
-            if (stack[i].getGroup() == TreeBuilder.TBODY_OR_THEAD_OR_TFOOT ||
-                    stack[i].getGroup() == TreeBuilder.TEMPLATE) {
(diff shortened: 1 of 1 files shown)


hsivonen (approved):
Good catch! Thanks!

hsivonen:
I pushed this via the command line as 6c124b3678b9d38a9cd3816b2699c54644cf132b and now I don't know how to make the GitHub PR metadata say merged instead of closed.

Issue #25, opened 2020-07-06
Create a new htmlparser-xom repository for the htmlparser.xom package?


carlosame:
PR #23 includes a modularization patch that has a dependency on XOM. Initially, the patch had a strict dependency on `nu.xom`, but as the result of a comment from @anthonyvdotbe I changed the requirement to `static` (_i.e._ optional). This is the most conservative thing to do right now but otherwise is formally wrong (because XOM objects are part of this library's public API) and defeats the purpose of a full modularization (the current setup is mostly equivalent to an automatic module name).

As a better solution, I suggest splitting the `htmlparser.xom` package to a different repository, and ship it as a different artifact (with different release cycles). As I believe that patches are better than words, I created that repository here:

https://github.com/carlosame/htmlparser-xom

and a `xom-removed` branch that removes the XOM support from `htmlparser` (it starts from my `validator-nu` branch that has the modularization patch):

https://github.com/carlosame/htmlparser/tree/xom-removed

A possible alternative would be to convert `htmlparser` to a modular Maven project (with a single Git repository but two artifacts being also created), however that would complicate the workflow for little to no benefit, and expose the project to a new category of Maven bugs (I mean bugs affecting modular builds only).

Thoughts?

carlosame:
It was decided against this in https://github.com/validator/htmlparser/issues/26#issuecomment-671834859, so closing.

Issue #14, opened 2020-04-22
Publish as modular JAR(s) to Maven Central


anthonyvdotbe:
Motivation:
* it's been since 2012 since the last release (there are more recent releases of the nu-validator branch, but that distribution [isn’t actually intended to be used outside the context of the HTML checker](https://github.com/validator/htmlparser/issues/11#issuecomment-612759158))
* this is a foundational library, so any library/application that uses it and wants to migrate to Java modules, relies on this library being modularized (though [automatic modules](https://github.com/validator/htmlparser/pull/13/) might help, they have their limits)

I'm willing to help by providing PRs such as:
* adapt the folder layout to Maven conventions
* split the project into multiple Maven modules as needed (since it's a lot easier if there's a 1-on-1 mapping between Maven and Java modules)
* make the Maven build work with a JDK 11+ release. In fact, such a JDK would be required in order to be able to compile the `module-info.java` file, so I'd like to upgrade the compiler level to 11 as well (the code can stay on the 1.5 syntactic level. I'm not planning to make any code changes, apart from package declarations and import statements)
* remove jchardet and XOM support: jchardet is an unmaintained library, and I don't see how XOM will be able to modularize itself any time soon, given that it relies on the `xml-apis:xml-apis` Maven artifact, and thus has a "split packages" issue

However, I have a few questions:
* first and foremost: if I provide PRs such as the above, will they be reviewed & accepted, and will a new release be published once modularization is done?
* there's a lot of files that are apparently unmaintained. Why aren't these deleted? It would help a lot if any legacy files would be deleted before starting
* there's 3 packages: encodings, htmlparser, saxtree. I understand why encodings is a separate package outside of htmlparser, but why is this also the case for saxtree? It's relied upon by the htmlparser package, and I don't readily see why someone using this library would use its classes directly
* following from my previous question: would it be acceptable to move saxtree into a package under the htmlparser package? Or is this package used directly by users of this library, and would changing the package name be a problem w.r.t. backward compatibility?

@hsivonen @sideshowbarker what are your thoughts on this?

sideshowbarker:
As far as Maven goes, this project so far isn’t internally relying on Maven for anything. Maven and the Central Repository aren’t the sole way or even the primary way that the HTML parser releases has been distributed. And Maven isn’t the primary build tool used for building releases.

Given that, making any disruptive changes to the sources primarily just to get things to work more easily with Maven seems like a “tail wagging the dog” kind of thing; in particular, the following:

[...]

PR #43, opened 2020-08-17
Prepare for 2.0 release

### LINKED ISSUE #26: Decide on the way forward
@hsivonen: as the de facto owner of this repository, I'd like to urge you to decide on the way forward, because I feel things are getting out of hand: @carlosame and I have different views on how to proceed, and we're both investing quite a bit of our spare time in implementing/defending our point of view. I believe that it's in everyone's interest for a final decision to be taken sooner rather than later. (I understand if modularization is not a priority for neither you nor your employer, but I believe it's not unreasonable to ask some of your time to settle this debate.)

## 1. Java modularization
There are plenty of options, starting from the most fine-grained modularization:

1. one module per XML API, with 2 base modules
   * `htmlparser.dom` -> `htmlparser.common`
   * `htmlparser.sax` -> `htmlparser.common`, `saxtree`
   * `htmlparser.xom` -> `htmlparser.common`
2. merge all htmlparser modules: `htmlparser` + `saxtree`
3. merge all modules into 1: `htmlparser`
4. merge all modules into 1, except the XOM module: `htmlparser` + `htmlparser.xom`
5. some other combination:
   * `htmlparser` + `htmlparser.xom` + `saxtree`
   * `htmlparser.jaxp` + `htmlparser.xom` + `htmlparser.common` + `saxtree`
   * ...

Q1: which combination makes most sense to you?

## 2. Project organization
Assuming there's at least 2 modules in the chose combination, there's multiple options on how to organize them:

1. a single multi-module Maven project in a single git repo
2. multiple single-module Maven projects in multiple git repos
3. some other organization

(since @carlosame [claimed](https://github.com/validator/htmlparser/issues/25#issue-651625290) that the first option would "complicate the workflow [...] and expose the project to a new category of Maven bugs": I claim the exact opposite, so please disregard any Maven-related concerns)

Q2: which option works best for you & @sideshowbarker?

## 3. upcoming releases
I'd like to propose that, no matter the answer to Q1 and Q2, version 1.5 is released with minimal changes, i.e. only changes that are required to resolve #17. In particular: no automatic module name, no changes to the dependency versions, ... Once 1.5 is published, we can then implement 1. and 2. as decided for a 2.0 release.

Q3: do you agree with this proposal?

Thanks in advance for reading & answering the 3 questions above, and thereby bringing back peace and quiet to this repository.

anthonyvdotbe:
Closes #26, #25, #14.

To build, a JDK 11+ is required. To build the Javadoc, a JDK 15+ is required (otherwise it fails on the `xom` module. There's already an RC available at http://jdk.java.net/15/, and the final release is due next month).

The commits should be self-explanatory.

W.r.t. the cleanup: which of these commits must be reverted, and what can effectively be removed (also considering that it's trivial to resurrect anything from a Git repo)?

W.r.t. "Make jchardet & ICU4J heuristics a no-op": `jchardet` is problematic. When relying on its automatic module name, Maven gives the following warning:

````
[WARNING] ****************************************************************************************************************************************
[WARNING] * Required filename-based automodules detected: [jchardet-1.0.jar]. Please don't publish this project to a public artifact repository! *
[WARNING] ****************************************************************************************************************************************
````

So I really believe we should eliminate the dependency altogether. The easiest way to do so, is by simply making the related heuristic a no-op. In fact, I believe the same should be done with the `ICU4J` dependency. The usage of optional Maven dependencies and `requires static` clauses are indications that the current situation isn't quite right.

The current commit is the minimal change required to eliminate the dependencies.

Now there are 3 options:
1. remove `Heuristics` altogether, given that
   * it's only used if: it's explicitly enabled, and the given `InputSource` doesn't wrap a `Reader`, and both of the basic sniffers (`Bom` & `Meta`) fail
   * it's trivial for users to do the encoding detection themselves, using any library of their choosing
   * none of the current sniffers are state of the art

2. use an SPI: `htmlparser` should introduce an interface `EncodingSniffer` and its `module-info.java` should say `uses EncodingSniffer;`. Then the current sniffers should be moved into their own Maven module(s), implement `EncodingSniffer`, and be made available as services to `htmlparser`

3. merge as is: this is no more than a behavioral regression (except for the removal of the sniffers, but I don't see how direct usage of these classes could possibly be justified), so we could just wait for someone to report it. Then once (if ever) it happens, one of the former 2 options can be implemented

What's your opinion?

### CODE REVIEW

File: src/nu/validator/saxtree/DocumentFragment.java [hunk 1]
```diff
@@ -37,7 +37,7 @@
      * The constructor.
      */
     public DocumentFragment() {
-        super(new LocatorImpl());
+        super(new Locator2Impl());
```
carlosame, line 40: `Locator2Impl` is not a drop-in replacement for `LocatorImpl`. At least, `getXMLVersion()` is going to return `null` with the former and `1.0` with the latter.

carlosame:
> The easiest way to do so, is by simply making the related heuristic a no-op

I thought that there was already an agreement about not introducing backward incompatibilities at this point.

> The usage of optional Maven dependencies and `requires static` clauses are indications that the current situation isn't quite right.

On ICU4J, the `requires` need not be `static`, and the Maven dependency could be marked as not optional. Saying "I'm going to use `setCheckingNormalization(boolean)` so I need to add ICU4J to the path" isn't exactly obvious or straight-forward.

carlosame (commented):
That looks like something used by the C++ translator (d6df8ad).

carlosame (commented):
The `htmlparser` project without the C++ translation? Have you talked with @hsivonen about that?

pom.xml
```diff
@@ -43,31 +43,31 @@
...
     that it currently does not do or do something better, you need to write
     the changes you need yourself and contribute a patch via
-    http://bugzilla.validator.nu/
+    https://github.com/validator/htmlparser/issues
   -->
   <developers>
     <developer>
       <id>hsivonen</id>
       <name>Henri Sivonen</name>
       <email>hsivonen@iki.fi</email>
```
carlosame: I never liked including emails in POM files. Otherwise, this patch looks good.

carlosame (commented):
Too bad that you did this after changing the repository layout, otherwise this could have been a good candidate to cherry-pick just now.

PR #44, opened 2020-08-17
Add test automation

sideshowbarker:
The changes in this pull-request branch enable us, for all pull requests and pushes, to automate testing of the parser against the html5lib-tests suite — including CI execution (using GitHub Actions) of the tests .

The changes include the following commits:

* 15955c02862815b4c131775e4d33cd47504c2c97 Add Html5libTest
* c1c10352061a8f9026a865995ebe4332ede5ecb0 Add html5lib-tests as a submodule
* ff468880d64187ca02213b7a495ac0a73cbdb22d Add GitHub Actions build.yml to run Maven in CI
* 419cae541fae5e3b4ca6caa1deec58564d69b157 Exit 1 for test harness if any tests fail
* 2dcd4da054e5c21a7ae59c77848be385cda7a6d6 Make Html5libTest handle double-escaped tests
* 27c5068fa07e81e89f943878a14f62bc56e1f50a Make Html5libTest only check .dat and .test files
* 500a9f5387f792055c5864a0eee6547c318a8cb1 Make EncodingTester usable in testing parsed state (from PR #49)
* 8d31a0b460c16f016c74bbaedbc5404aa3b59f1c Make HtmlInputStreamReader sniffing limit settable (from PR #49)

…along with some related minor changes to get things working smoothly.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -80,8 +80,8 @@
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-compiler-plugin</artifactId>
         <configuration>
-          <source>1.5</source>
-          <target>1.5</target>
+          <source>1.8</source>
+          <target>1.8</target>
         </configuration>
       </plugin>
       <plugin>
@@ -136,7 +136,11 @@
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-surefire-plugin</artifactId>
         <configuration>
-          <skip>true</skip>
+          <test>Html5libTest</test>
+          <testFailureIgnore>true</testFailureIgnore> <!-- FIXME: Remove this testFailureIgnore after we have all tests passing -->
+          <additionalClasspathElements>
+            <additionalClasspathElement>${project.build.testSourceDirectory}/test/resources</additionalClasspathElement>
+          </additionalClasspathElements>
         </configuration>
       </plugin>
       <plugin>
--- a/pom.xml
+++ b/pom.xml
@@ -115,8 +115,12 @@
                 <property name="translator.sources" value="${basedir}/translator-src"/>
                 <property name="translator.classes" value="${project.build.directory}/translator-classes"/>
                 <mkdir dir="${translator.classes}"/>
-                <javac srcdir="${translator.sources}" includes="nu/validator/htmlparser/generator/ApplyHotSpotWorkaround.java" destdir="${translator.classes}" includeantruntime="false"/>
-                <java classname="nu.validator.htmlparser.generator.ApplyHotSpotWorkaround">
+                <javac srcdir="${translator.sources}" destdir="${translator.classes}"
+                       includes="nu/validator/htmlparser/generator/ApplyHotSpotWorkaround.java"
+                       includeantruntime="false"
+                       fork="${fork}"/>
+                <java classname="nu.validator.htmlparser.generator.ApplyHotSpotWorkaround"
+                      fork="${fork}">
                   <classpath>
                     <pathelement location="${translator.classes}"/>
                   </classpath>
@@ -227,6 +231,7 @@
     <rpm.java.dir>/usr/share/java</rpm.java.dir>
     <rpm.javadoc.dir>/usr/share/javadoc</rpm.javadoc.dir>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+    <fork>false</fork>
   </properties>
   <profiles>
     <profile>
@@ -234,6 +239,9 @@
       <activation>
         <jdk>1.8</jdk>
       </activation>
+      <properties>
+        <fork>true</fork>
+      </properties>
       <dependencies>
         <dependency>
           <groupId>com.sun</groupId>
--- a/pom.xml
+++ b/pom.xml
@@ -87,15 +87,6 @@
       <plugin>
         <artifactId>maven-antrun-plugin</artifactId>
         <version>1.7</version>
-        <dependencies>
-          <dependency>
-            <groupId>com.sun</groupId>
-            <artifactId>tools</artifactId>
-            <version>1.5.0</version>
-            <scope>system</scope>
-            <systemPath>${java.home}/../lib/tools.jar</systemPath>
-          </dependency>
-        </dependencies>
         <executions>
           <execution>
             <id>intitialize-sources</id>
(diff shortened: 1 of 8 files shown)


pom.xml
```diff
@@ -104,6 +104,45 @@
...
           </execution>
+          <execution>
+            <id>test</id>
+            <phase>test</phase>
+            <goals>
+              <goal>run</goal>
+            </goals>
+            <configuration>
+              <target>
+                <property name="test_classpath" refid="maven.test.classpath"/>
```
carlosame, line 115: I suggest adding `<property name="build.compiler" value="extJavac"/>` here and checking whether `fork="true"` is still required.

sideshowbarker: > I suggest adding `<property name="build.compiler" value="extJavac"/>` here and checking whether `fork="true"` is still required.

Thanks — tried adding that just now, and removing `fork="true"`, but I get the following error:

> An Ant BuildException has occured: java.lang.NoClassDefFoundError: jdk/internal/reflect/ConstructorAccessorImpl

sideshowbarker:
So as y’all may have noticed, I changed the code in this PR branch to remove the kludgey AntRun additions I’d made, and instead use @anthonyvdotbe’s `Html5libTest` class.

That approach currently requires adding an extra config setting, to teach Surefire where the `html5lib-tests` directory is.

That config setting would be unnecessary if/when we end up moving to a new directory layout that follows Maven conventions. But after we do ever get there, the POM can be updated to drop the extra config setting.

@anthonyvdotbe Can you confirm you’re agreeing to license the `Html5libTest` code under the existing project license in the https://github.com/validator/htmlparser/blob/master/LICENSE.txt file? (It’s essentially the MIT license.)

If so, do you want to have a `Copyright (c) 2020 Anthony Vanelverdinghe` line included in the license/copyright header comment in the source — along with a `Copyright (c) 2020 Mozilla Foundation` line?

Or are you instead OK with the license header having only the `Copyright (c) 2020 Mozilla Foundation` line? (In other words, in that case, could you confirm that you’re waiving your right to have your name included in the copyright statement.)

anthonyvdotbe:
> @anthonyvdotbe Can you confirm you’re agreeing to license the `Html5libTest` code under the existing project license in the https://github.com/validator/htmlparser/blob/master/LICENSE.txt file? (It’s essentially the MIT license.)

Yes, I confirm that I agree.

> Or are you instead OK with the license header having only the `Copyright (c) 2020 Mozilla Foundation` line? (In other words, in that case, could you confirm that you’re waiving your right to have your name included in the copyright statement.)

I confirm that I'm waiving my right to have my name included in the copyright statement.

carlosame:
> I changed the code in this PR branch to remove the kludgey AntRun additions I’d made, and instead use @anthonyvdotbe’s `Html5libTest` class.

The conversion of all the tests (including `Html5libTest`) to JUnit is something that could be done after this PR is merged, and mostly looks like a search-and-replace trick.

sideshowbarker:
> The conversion of all the tests (including `Html5libTest`) to JUnit is something that could be done after this PR is merged, and mostly looks like a search-and-replace trick.

OK. But I personally know almost nothing about JUnit, so anything that’d need to be done to move to JUnit would have to come in form of PRs with patches — all of which I’d be happy to test, and to review (and learn from…).

But if we were to move to JUnit, I would hope there’d be some concrete measurable benefit to doing so. I mean, the current POJO way that `Html5libTest` uses seems to me to at least have the advantage of getting the job done while also being very simple. I would imagine that moving to JUnit would make it less simple. So I’d wonder what concretely we’d be gaining by losing the simplicity and taking on the cost of some greater complexity.

Or maybe if I had more domain knowledge in this area, it’d be clear to me why moving to JUnit makes the best sense. Maybe  there’s some intrinsic/potential value in it that’s generally well understood already by most people.

carlosame:
> Maybe there’s some intrinsic/potential value in it that’s generally well understood already by most people.

For example, not having to tell the Surefire configuration which files are tests and which aren't.

> I personally know almost nothing about JUnit

It is easy and fast to learn the basics.

> so anything that’d need to be done to move to JUnit would have to come in form of PRs with patches

I do not have plans to send more PRs.

carlosame:
I see that you re-enabled testing for Java 8, which is a good thing (commits 05b9024 and a27c7b6), but using Java 8 is unlikely to get along very well with the current modularization patches (as long as there is the requirement that the minimum JDK to process the POM is 11). And running automated testing with Java 11 or higher is going to be a quite different game once the project is modularized.

It's not that you cannot do either of those things (you can even use Java 8 if the POM is complicated enough) but modularization is highly coupled with testing, and setting up an automated testing infrastructure without accounting for modules has the potential of being a waste of time.

IMHO either this PR or the modularization PR should be merged as soon as possible (BTW after making any necessary changes), and then the other PR can rebase on top of it. Yes I assume that you can't do a lot either way, just making sure that you are aware of this.

sideshowbarker:
@carlosame Thanks for the heads-up about the issue dealing with Java 8 in the face of the modularization patches.

It’d be pretty disappointing if Maven, in all its baroque complexity, didn’t provide a way to build a modularized jar for current Java versions while also remaining able to ensure the code at least continues to compile under Java 8.

I personally am not willing to cavalierly ignore all the users who possibly might be constrained to needing to use the parser in a Java 8 environment. I don’t have a way to measure how many such users the parser might have. For all I know, it could be there are so few users of the parser that need Java 8 compatibility that we don’t need to worry about it. Or it could be that there are some important users for whom we’d be abandoning when we rightly shouldn’t be.

But if we can’t even ourselves continue to have an automated way to test that the code continues compiles under Java 8, then we have no way to know when some code change we made has broken compatibility with Java 8.

sideshowbarker:
On further reflection, it occurs to me to ask: As far as building with Maven goes, couldn’t we just have one pom.xml for building/testing the “normal” build for Java 11+, and then a second pom.xml for building/testing the Java 8 build?

carlosame:
> ensure the code at least continues to compile under Java 8

Hmm that's not what I was meaning. The modularization patches in PR #43 generate bytecode for Java 8 (IMHO should be 7) except for the `module-info` file. One thing is "compiling for Java 8", another is "testing with Java 8". In fact, the most likely outcome for this project's modularization is that the actual bytecode being tested will be version 8 but the JDK used during the tests is 11, 15 or whatever.

Actually, your current patches in this PR are always testing Java 8 bytecode, you are just producing and testing that level-8 bytecode with different JDKs. And source-level compatibility is set to 8 as well.

carlosame:
Additional clarification to my previous post: the idea is that Java 8 users will be able to use the compiled _jar_ files, but won't be able to build them. A modular JDK is required to build.

If you still want to test with Java 8 and want to avoid undesirable POM hacks, you could set up a separate test project that would depend on a level-8 [_test-jar_](https://maven.apache.org/guides/mini/guide-attached-tests.html) produced by the main build. But I see no real reason to do that.

sideshowbarker:
I believe everything in this PR branch is now also in the main branch — so I’m going ahead and closing this.

PR #48, opened 2020-08-20
Conform encoding-label matching to Encoding spec

sideshowbarker:
This change makes the parser’s encoding-name matching conform to the current Encoding spec at https://encoding.spec.whatwg.org/#concept-encoding-get — which requires that only leading and trailing whitespace be removed from a string before checking if it matches any valid encoding names.

Otherwise, without this change, the parser instead implements https://www.unicode.org/reports/tr22/tr22-8.html#Charset_Alias_Matching — which requires deleting “all characters except a-z, A-Z, and 0-9” from a string before checking if it matches any valid encoding names. That difference makes us fail two html5lib-tests cases.

---

The html5lib-tests cases that we fail without this change are the following:

* https://github.com/html5lib/html5lib-tests/blob/master/encoding/tests1.dat#L65
* https://github.com/html5lib/html5lib-tests/blob/master/encoding/tests1.dat#L162


src/nu/validator/htmlparser/io/Encoding.java
```diff
@@ -254,9 +254,7 @@ public static String toNameKey(String str) {
             if (c >= 'A' && c <= 'Z') {
                 c += 0x20;
             }
-            if (!((c >= '\t' && c <= '\r') || (c >= '\u0020' && c <= '\u002F')
-                    || (c >= '\u003A' && c <= '\u0040')
-                    || (c >= '\u005B' && c <= '\u0060') || (c >= '\u007B' && c <= '\u007E'))) {
+            if (!Arrays.asList('\t','\n','\f','\r','\u0020').contains(c)) {
```
hsivonen, line 257: The performance characteristics of `Arrays.asList` are a bit unclear to me, but it seems bad to call it again on each loop iteration. I think `(c == ' ' || c == '\t' || c == '\n' || c == '\f' || c == '\r')` would probably be the best thing to do here.

hsivonen (changes requested):

src/nu/validator/htmlparser/io/Encoding.java
sideshowbarker: > The performance characteristics of `Arrays.asList` are a bit unclear to me, but it seems bad to call it again on each loop iteration. I think `(c == ' ' || c == '\t' || c == '\n' || c == '\f' || c == '\r')` would probably be the best thing to do here.

Thanks, yeah, I should have just done it that way to begin with. Anyway, changed it to that now. Amended to the existing commit, and force-pushed to the branch — so that we just have one commit here, and so need to squash before merging.

sideshowbarker:
I’m seeing some possible regressions in the HTML checker as a result of this —
```
error: The encoding “utf-8” is not an IANA-registered encoding and did not use the “x-” prefix. (Charmod C023)
info warning: The character encoding “utf-8” is not widely supported. Better interoperability may be achieved by using “UTF-8”.
info warning: Using “x-iso-2022-cn-gb” instead of the declared encoding “utf-8”.
```
That is, this change seems to have caused the parser to fail to parse `utf-8` case-insensitively.

And then for some reason, the parser seems to randomly picking some other encoding to use — the encoding name in the last error message above is non-deterministic. When I re-run the tests, different encoding names show up:
```
info warning: Using “x-machebrew” instead of the declared encoding “utf-8”.
```
or:
```
info warning: Using “x-maccroatian” instead of the declared encoding “utf-8”.
```
… etc.

sideshowbarker:
d’oh — never mind my previous comment; I see now that I just neglected to negate the condition. Will push an update

sideshowbarker:
OK, I found that the existing code had effectively been using, as the internal canonical name, names with all dashes and underscores removed. That results in a conformance violation, because it means that, for example, we’d treat the string `ut-f8` (invalid as an encoding name) the same as `utf-8` or `utf8`.

So with 92275d9, I’ve updated the checking to conform to exactly what the Encoding spec requires — which is to do an exact comparison of the lower-cased input string to the set of labels in https://encoding.spec.whatwg.org/#names-and-labels

src/nu/validator/htmlparser/io/Encoding.java
```diff
@@ -106,11 +329,12 @@
             Charset cs = entry.getValue();
             String name = toNameKey(cs.name());
             String canonName = toAsciiLowerCase(cs.name());
-            if (!isBanned(name)) {
+            if (!isBanned(stripDashAndUnderscore(name))) {
```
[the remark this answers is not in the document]
sideshowbarker: > It's a bit unclear to me where this patch is presently going in terms of compliance if `isBanned` stays.

I guess I see this as another case where (with this patch applied) the implementation is functionally equivalent to the spec prose. But I agree it would be better to make it as explicitly more aligned with the spec prose as possible — for example, by rewriting the code to have an allow list rather than a blocklist.

So I can make some time to work on the code in this branch do make it like that.

sideshowbarker:
OK, I’ve now pushed a large number of commits to this branch that together make the parser’s handling of encodings conformant with current requirements in the Encoding spec and HTML spec. The key changes are the following:

81c40ef Conform MetaSniffer "x-user-defined" handling
f9b3796 Sync w/ “Changing the encoding while parsing” algo
ceb9fe3 Drop UTF-32 & “UTF-16”; use UTF-16BE and UTF-16LE
9d2d39d Drop Encoding.toAsciiLowerCase; use toLowerCase()
beb060b Conform encoding-label matching to Encoding spec
8c62d2d Conform supported encodings to Encoding spec
5cfabbe Drop superseded encoding-checking code
343d86f Drop “actual HTML encoding”-related code
03954fb Drop/replace “is ASCII superset”-checking code

The branch also includes some related minor changes to get things working smoothly.

I broke the changes out into separate commits in order to facilitate easier review.

So in reviewing this, you probably want to be starting from the https://github.com/validator/htmlparser/pull/48/commits view rather than the https://github.com/validator/htmlparser/pull/48/files view.

sideshowbarker:
OK, I’m now a bit embarrassed to admit: it hadn’t dawned on me until now that the changes to the `Encoding` class in this patch are completely superseded by the code in the `encodings` branch of the repo.

So I’m now wondering whether we should just not spend any more time at all on this patch but instead try to get the code on the `encodings` branch finished and ready to merge to the main branch and to release.

I’m certainly willing to put time into helping with that.

carlosame:
> get the code on the `encodings` branch finished and ready to merge

My understanding was that `encodings` is an archive branch kept for reference, not initially intended for merging back. Perhaps I'm missing something.

PR #56, opened 2021-10-07
increase bufferSpace by at least 25%

MarkusKull:
to remedy an O(n^2) performance problem when parsing e.g. inline images.

The performance problem is noticeable for e.g. `<img src=\"data:image/png;base64,iVBORw0KGgoAAAA...` when the image has a size of  multiple megabytes. An incremental increase of some KBs each does not help much, thus this PR increases by a proportional factor of 25%.


sideshowbarker:
> Seems reasonable enough to me. @sideshowbarker ?

Yep, sounds good to me

PR #59, opened 2021-10-13
chore: setup Dependabot

nschonni:


.github/workflows/build.yml
```diff
@@ -1,6 +1,10 @@
 name: Build
 
-on: [push, pull_request]
+on:
+  push:
+    branches:
+      - main
```
nschonni, line 6: Noticed that the default branch hasn't been renamed here, so this won't currently build on any pushes

PR #64, opened 2021-12-08
Mozilla bug 1701828 addendum - More comments for meta charset rewrite…

hsivonen:
… for Gecko.

Differential Revision: https://phabricator.services.mozilla.com/D125808

### DIFF
--- a/src/nu/validator/htmlparser/impl/Tokenizer.java
+++ b/src/nu/validator/htmlparser/impl/Tokenizer.java
@@ -3181,6 +3181,12 @@ public class Tokenizer implements Locator, Locator2 {
                             case '>':
                                 cstart = pos + 1;
                                 state = transition(state, Tokenizer.DATA, reconsume, pos);
+                                // Since a CDATA section starts with a less-than sign, it
+                                // is participates in the suspension-after-current-token
+                                // behavior. (The suspension can be requested when the
+                                // less-than sign has been seen but we don't yet know the
+                                // resulting token type.) Therefore, we need to deal with
+                                // a potential request here.
                                 suspendIfRequestedAfterCurrentNonTextToken();
                                 if (shouldSuspend) {
                                     break stateloop;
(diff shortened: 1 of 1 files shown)


src/nu/validator/htmlparser/impl/Tokenizer.java
```diff
@@ -3181,6 +3181,12 @@ private void ensureBufferSpace(int inputLength) throws SAXException {
                             case '>':
                                 cstart = pos + 1;
                                 state = transition(state, Tokenizer.DATA, reconsume, pos);
+                                // Since a CDATA section starts with a less-than sign, it
+                                // is participates in the suspension-after-current-token
+                                // behavior. (The suspension can be requested when the
+                                // less-than sign has been seen but we don't yet know the
+                                // resulting token type.) Therefore, we need to deal with
+                                // a potential request here.
                                 suspendIfRequestedAfterCurrentNonTextToken();
```
sideshowbarker, line 3190:
```suggestion
                                // Since a CDATA section starts with a less-than sign, it
                                // participates in the suspension-after-current-token
                                // behavior. (The suspension can be requested when the
                                // less-than sign has been seen but we don't yet know the
                                // resulting token type.) Therefore, we need to deal with
                                // a potential request here.
                                suspendIfRequestedAfterCurrentNonTextToken();
```

src/nu/validator/htmlparser/impl/Tokenizer.java
```diff
@@ -6189,6 +6195,19 @@ private void ensureBufferSpace(int inputLength) throws SAXException {
...
+                            // Since XML View Source doesn't use the
+                            // suspension-after-current-token facility, its extension
+                            // to processing instruction states is strictly unnecessary
+                            // at the moment. However, if these states ever were to be
+                            // used together with the suspension-after-current-token
+                            // facility, these states would need to participate, since
+                            // suspension could be requested when only less-than has been
+                            // seen and we don't yet know if we end up here. Handling
+                            // the currently unnecessary case in order to avoid leaving
+                            // a trap for future modification.
```
sideshowbarker, line 6210:
```suggestion
                            // Processing instruction syntax goes through these
                            // states only in Gecko's XML View Source--not in HTML
                            // parsing in Java or in Gecko.
                            // Since XML View Source doesn't use the
                            // suspension-after-current-token facility, its extension
                            // to processing-instruction states is strictly unnecessary
                            // at the moment. However, if these states ever were to be
                            // used together with the suspension-after-current-token
                            // facility, these states would need to participate, since
                            // suspension could be requested when only less-than has been
                            // seen and we don't yet know if we end up here. Handling
                            // the currently-unnecessary case in order to avoid leaving
                            // a trap for future modification.
```

src/nu/validator/htmlparser/impl/Tokenizer.java
```diff
@@ -6959,6 +6978,11 @@ private void emitDoctypeToken(int pos) throws SAXException {
         suspendIfRequestedAfterCurrentNonTextToken();
     }
 
+    /**
+     * If a previous call to <code>suspendAfterCurrentTokenIfNotInText()</code>
+     * happened in a non-text context, this method turns that deferred suspension
+     * request into an immediately pending suspension request.
+     */
```
sideshowbarker, line 6985:
```suggestion
    /**
     * If a previous call to <code>suspendAfterCurrentTokenIfNotInText()</code>
     * happened in a non-text context, this method turns that deferred suspension
     * request into an immediately-pending suspension request.
     */
```

sideshowbarker (approved):

hsivonen:
Thanks!

PR #70, opened 2022-06-22
Mozilla bug 1372276 - Remove menuitem.

emilio:


emilio:
r? @hsivonen

I don't think the build failure is related to my patch, but let me know if that's wrong.

hsivonen:
The failure relates to a mismatch of the JSON library version. It seems that it's [my fault](https://github.com/validator/htmlparser/commit/0d330112715f5d1bc8b206c08d2a64a2aab6fd82).

hsivonen:
I reverted the change from my previous changeset that shouldn't have been there in the first place. It seems, though, that restarting the tests use the base revision that the PR was made from and not the new trunk.

emilio:
Rebased on top of current trunk.

hsivonen:
Oops. This needs hash regeneration still.

PR #74, opened 2022-09-27
Move the self-closing-tag warning to the TreeBuilder code

sideshowbarker:


src/nu/validator/htmlparser/impl/TreeBuilder.java
```diff
@@ -438,6 +438,8 @@
 
     private boolean reportingDoctype = true;
 
+    private HashMap<String, String> errorProfileMap = null;
```
hsivonen, line 441: Is this necessary? Can this be obtained from the tokenizer (even though it only exists on a subclass of the tokenizer)?

src/nu/validator/htmlparser/impl/TreeBuilder.java
```diff
@@ -1458,6 +1460,17 @@ public final void startTag(ElementName elementName,
         flushCharacters();
 
         // [NOCPP[
+        if (contextNamespace == "http://www.w3.org/1999/xhtml" &&
```
hsivonen, line 1463: `contextNamespace` is the namespace of the fragment parsing context node, so it's the wrong thing here.

src/nu/validator/htmlparser/impl/TreeBuilder.java
```diff
@@ -1458,6 +1460,17 @@ public final void startTag(ElementName elementName,
...
+                selfClosing && errorProfileMap != null &&
+                errorProfileMap.get("html-strict") != null) {
+            warn("Self-closing tag syntax in text/html documents is widely"
+                    + " discouraged; it’s unnecessary and interacts badly"
+                    + " with other HTML features (e.g., unquoted attribute"
+                    + " values). If you’re using a tool that injects"
+                    + " self-closing tag syntax into all void elements,"
+                    + " without any option to prevent it from doing so,"
+                    + " then consider switching to a different tool.");
+        }
```
hsivonen, line 1473: This now emits a duplicate message when there is an actual error here:
https://searchfox.org/mozilla-central/rev/929b2a7154e463674ebac497d4a89208eec1b8f7/parser/html/javasrc/TreeBuilder.java#2846

To get this right, I think you need two booleans initialized at to top of the method in a `NOCPP` block: `boolean wasSelfClosing = selfClosing;` and `boolean voidElement = false;` Then the non-foreign cases that set `selfClosing = false;` should, in a `NOCPP` block also set `voidElement = true;`. This should then be enough to decide if a non-error is appropriate as an `else` branch to the check for `selfClosing` at the end of the method.

hsivonen (changes requested):

hsivonen (changes requested):

src/nu/validator/htmlparser/impl/TreeBuilder.java
sideshowbarker:
```java
    private HashMap<String, String> errorProfileMap = null;
```

> Is this necessary? Can this be obtained from the tokenizer (even though it only exists on a subclass of the tokenizer)?

I added a `getErrorProfile()` method to the tokenizer. But I don’t see how to make it to compiled if it’s only a method of the ErrorReportingTokenizer subclass — so I also made it a method of the Tokenizer superclass, and then overrode it in the ErrorReportingTokenizer subclass.

Is there some way I could instead make it work if it only exists in the ErrorReportingTokenizer subclass (so that I don’t have to also clutter the superclass with it)?

Also, did you have in mind that a getter isn’t really needed here, but instead I could just make the errorProfileMap of the tokenizer protected (instead of private), and then just have the treebuilder code use that (rather then needing a getter)?

src/nu/validator/htmlparser/impl/TreeBuilder.java
sideshowbarker: > `contextNamespace` is the namespace of the fragment parsing context node, so it's the wrong thing here.

Yeah, I subsequently also realized that after testing and looking through the code more…

Anyway, that’s gone now — since we now just set the voidElement flage per-element (avoiding the foreign elements)

src/nu/validator/htmlparser/impl/TreeBuilder.java
sideshowbarker: > To get this right, I think you need two booleans initialized at to top of the method in a `NOCPP` block: `boolean wasSelfClosing = selfClosing;` and `boolean voidElement = false;` Then the non-foreign cases that set `selfClosing = false;` should, in a `NOCPP` block also set `voidElement = true;`. This should then be enough to decide if a non-error is appropriate as an `else` branch to the check for `selfClosing` at the end of the method.

Thanks, did it that way

src/nu/validator/htmlparser/impl/TreeBuilder.java
```diff
@@ -2857,6 +2885,16 @@ public final void startTag(ElementName elementName,
         }
         if (selfClosing) {
             errSelfClosing();
+        } else if (wasSelfClosing && voidElement
```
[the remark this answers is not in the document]
sideshowbarker: > Please put the `else` part in a `NOCPP` block.

OK, just push a change that does that. Please double-check to make sure I got that part right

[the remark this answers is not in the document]
hsivonen: Thanks.

hsivonen (approved):

Issue #82, opened 2023-05-25
Ambiguous ampersands are not detected


ezequiel-garzon:
According to [the HTML standard](https://html.spec.whatwg.org/multipage/syntax.html#syntax-ambiguous-ampersand), "an ambiguous ampersand is a U+0026 AMPERSAND character (&) that is followed by one or more ASCII alphanumerics, followed by a U+003B SEMICOLON character (;), where these characters do not match any of the names given in the named character references section".

Maybe I'm missing something, but shouldn't then something like `&ThisAmpersandShouldBeDeemedAmbiguous;` raise an error, or a warning? I know it used to, but I've checked both on a Mac with version 23.4.11, as well as on <https://validator.w3.org/>, and no error or warning is now raised. Thanks in advance.

sideshowbarker:
@ezequiel-garzon Thanks much for catching it. It was a great catch because, embarrassingly, it appears we’ve unfortunately had this bug for almost two years now — and the effect of it is that for those almost two years now, the HTML checker hasn’t been reporting any errors for almost all cases of invalid named character references.

In other words, when, for example, people have accidentally made minor spelling mistakes to otherwise-valid named character references, the HTML checker hasn’t been catching that and reporting it so that they can fix their spelling mistakes.

I’ve fixed this in a feature branch with https://github.com/validator/htmlparser/pull/83 — and for now, I’ve switched the HTML checker to being built from that branch, and pushed the updates to https://validator.w3.org/nu/

But I’ll keep this issue open until the fix gets merged into the main branch of the HTML parser code here.

ezequiel-garzon:
My pleasure, @sideshowbarker. Thank _you_ for taking care of this and so many other projects. I checked many, many times before reporting as I thought I was doing something wrong.

PR #83, opened 2023-05-29
Fix handling of "ambiguous ampersands" to report error when expected

sideshowbarker:
Fixes https://github.com/validator/htmlparser/issues/82. 44427c0df60ec899c52b149bba03e1abffad2f5e broke handling of "ambiguous ampersands" for the case where we've already examined enough of string after the ampersand to conclude that the substring we have so far doesn't match the beginning of any known named character reference.

This change re-conforms the error-reporting for that case to the requirements in the HTML spec. Otherwise, without this change, no error is reported as expected in many or most cases where an ampersand doesn't actually start a character reference.


nschonni:
@dill0wn that looks like a regression to me. I don't think it should be flagging the query string parameters

sideshowbarker:
> @dill0wn that looks like a regression to me. I don't think it should be flagging the query string parameters

Yeah, it‘s a regression. See also https://github.com/validator/validator/issues/1595

sideshowbarker:
@dill0wn
> I just would like to double check that this is bad practice and something we should fix in our code. Because it seems pretty innocuous.

Per the current spec, it’s not bad practice and not something you need to fix in your code in order for it to be valid.

We changed the parser algorithm in the spec years ago to allow for ampersands in URLs in attribute values, while still disallowing it in other cases where we wanted to disallow it — and then we attempted to change the parser to match the behavior required by the spec. It seems now we got the allow-in-attribute-values behavior done right in the parser, but as a result, somewhere along the lines, we seem to have ended up allowing ampersands everywhere. So we still need to fix that. But in the meantime, I need to revert this change.

sideshowbarker:
I’ve fixed this in the HTML checker sources and pushed the fix to https://validator.w3.org/nu/

PR #86, opened 2023-09-21
Remove unused interner argument from elementNameByBuffer.

mstange:


mstange:
No there isn't; I wasn't sure about the process of writing patches for the HTML parser so this was mostly a test PR to see what it was like :)

Should I file a Gecko bug?

hsivonen:
> No there isn't; I wasn't sure about the process of writing patches for the HTML parser so this was mostly a test PR to see what it was like :)
> 
> Should I file a Gecko bug?

Gecko developers usually start from a Gecko bug, but I can treat this as a non-Gecko contribution, in which case I'll file a bug to land the regenerated code.

(Looks like I failed to remove this argument in https://bugzilla.mozilla.org/show_bug.cgi?id=1355769 )

I'll merge this after another in-flight PR merges so that the m-c landing order and the landing order here stay in sync.

Thanks!

hsivonen (approved):

hsivonen:
Filed https://bugzilla.mozilla.org/show_bug.cgi?id=1861232

mstange:
> Gecko developers usually start from a Gecko bug

Is this process documented somewhere?

hsivonen:
> > Gecko developers usually start from a Gecko bug
> 
> Is this process documented somewhere?

I think it isn't. It's mostly about Gecko developers usually seeing some problem from the Gecko perspective and, therefore, starting with Bugzilla and Phabricator.

PR #88, opened 2023-10-18
Mozilla bug 1857221 - Avoid tracking the line and column number when parsing innerHTML.

hsivonen:
Differential Revision: https://phabricator.services.mozilla.com/D191270

### DIFF
--- a/translator-src/nu/validator/htmlparser/cpptranslate/CppVisitor.java
+++ b/translator-src/nu/validator/htmlparser/cpptranslate/CppVisitor.java
@@ -439,6 +439,16 @@ public class CppVisitor extends AnnotationHelperVisitor<LocalSymbolTable> {
         printer.print(className);
         printer.printLn(".h\"");
         printer.printLn();
+        
+        if ("Tokenizer".equals(javaClassName)) {
+            String loopPolicyInclude = cppTypes.loopPolicyInclude();
+            if (loopPolicyInclude != null) {
+                printer.print("#include \"");
+                printer.print(loopPolicyInclude);
+                printer.printLn(".h\"");
+                printer.printLn();                
+            }
+        }
     }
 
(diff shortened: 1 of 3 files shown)


translator-src/nu/validator/htmlparser/cpptranslate/CppVisitor.java
```diff
@@ -1364,6 +1383,10 @@ public void visit(MethodCallExpr n, LocalSymbolTable arg) {
                 printer.print(cppTypes.stateLoopPolicies()[stateLoopCallCount]);
                 printer.print(">");
                 stateLoopCallCount++;
+            } else if (cppTypes.requiresTemplateParameter(n.getName())
+                    && "Tokenizer".equals(javaClassName)
```
smaug----, line 1387: In my mind this would be easier to read if one would first check class name and then effectively the method name. But looks like you're just following the existing style, so fine.

translator-src/nu/validator/htmlparser/cpptranslate/CppVisitor.java
```diff
@@ -1280,6 +1290,15 @@ public void visit(MethodCallExpr n, LocalSymbolTable arg) {
         } else if ("transition".equals(n.getName())
                 && n.getScope() == null) {
             visitTransition(n, arg);
+        } else if ("checkChar".equals(n.getName())
+                && n.getScope() == null) {
```
smaug----, line 1294: Does null scope mean static method?

smaug---- (approved):

translator-src/nu/validator/htmlparser/cpptranslate/CppVisitor.java
hsivonen: I don't actually recall with 100% certainty, since I copypasted from existing code, but I believe it means that the call isn't qualified by an expression followed by a dot.

Thanks for the review.

PR #89, opened 2023-10-25
Mozilla bug 1861010 - Improve nsHTMLAttributeName performance for non-interned attributes.

emilio:


hsivonen (approved):
LGTM. Thank you!

I'll merge after the Gecko side lands.

PR #91, opened 2024-02-20
Prevent NPE in HtmlAttributes.getIndex

sideshowbarker:
Apparently it’s possible for a `mode` value to be such that `AttributeName.getQName(mode)` returns null. So this change adds a null check to catch that.

Otherwise, without this change, some documents cause that code to throw a NullPointerException.

The NullPointerException is reproducible by parsing the document at https://usbliss.com/?attachment_id=2232 — or by giving that document to the https://validator.w3.org/nu/ checker to check:

https://validator.w3.org/nu/?doc=https://usbliss.com/?attachment_id=2232

That causes the checker to emit the following message:

> Internal Error: Oops. That was not supposed to happen. A bug manifested itself in the application internals. Unable to continue. Sorry. The admin was notified.

(The admin who gets notified is me, and notification I get is a stack trace pointing to the NPE inHtmlAttributes.getIndex.)

### DIFF
--- a/src/nu/validator/htmlparser/impl/HtmlAttributes.java
+++ b/src/nu/validator/htmlparser/impl/HtmlAttributes.java
@@ -184,7 +184,8 @@ public final class HtmlAttributes implements Attributes {
 
     public int getIndex(String qName) {
         for (int i = 0; i < length; i++) {
-            if (names[i].getQName(mode).equals(qName)) {
+            if (names[i].getQName(mode) != null &&
+                    names[i].getQName(mode).equals(qName)) {
                 return i;
             }
         }


hsivonen:
That's really weird. I guess I will need to see in a debugger what's going on.

sideshowbarker:
The HtmlAttributes.getIndex NPE is getting thrown in the production by the W3C service every 5–10 minutes. And I get a notification every time. So I’ve created a https://github.com/validator/htmlparser/tree/vnu branch for the htmlparser repo and pushed the workaround from here to that branch — and in  https://github.com/validator/validator/commit/7d62ce104be8319ff3cd01fc904bc594afde4010 I’ve switched the checker code over for now to using that htmlparser branch.

sideshowbarker:
OK, looking back through my notifications and logs for the documents that are causing the NPE to be thrown — and then manually running those documents back through the checker myself, I find that one thing they all so far appear to have in common is that in every case they’re also causing a _“Stray start tag `html`”_ error to be reported.

So, I can imagine that what may be happening is that the parser is trying to check the attributes on that stray `html` start tag, but the state is such that `AttributeName.getQName(mode)` is unexpectedly null in that case.

sideshowbarker:
OK, the document `<html><html id>` is a reduced case with which the NPE/regression can be reproduced. Or with most any other arbitrary attribute name in that second `html` start tag.

Some bisecting seems to indicate the change in https://github.com/validator/htmlparser/commit/32c2256f33dabf071de76ad335b606614d0ad42a introduced the regression for that case. If I build instead from https://github.com/validator/htmlparser/commit/43aa7fcb6ac971c025bb7f80717ea9695c9a074a — the parent of that change — the NPE doesn’t get thrown.

Looking briefly through the https://github.com/validator/htmlparser/commit/32c2256f33dabf071de76ad335b606614d0ad42a change, nothing obvious jumps out to me as being something that’d lead to the behavior breaking for the `<html><html id>` case. But I’ll make some time to try digging into it further.

cc @emilio

emilio:
@sideshowbarker can you paste the full stack trace?

emilio:
@sideshowbarker I think the right fix is #92, but if you could confirm it'd be appreciated.

PR #113, opened 2025-12-30
Implement <select> parser “relaxation” — for “customizable” <select>

sideshowbarker:
Relax `select` parsing to allow more child elements, and handle the `selectedcontent` element for cloning `option` content.


hsivonen:
Sorry I missed this over the holidays.

sideshowbarker:
> Sorry I missed this over the holidays.

No worries — there's no real rush on it

src/nu/validator/htmlparser/sax/SAXTreeBuilder.java
```diff
@@ -197,4 +198,52 @@ private Node previousSibling(Node table) {
...
+                // Recursively clone children
+                Node child = srcElem.getFirstChild();
+                while (child != null) {
+                    deepCloneNode(child, cloneElem);
+                    child = child.getNextSibling();
+                }
+                break;
+            case CHARACTERS:
+                // Clone the characters
+                String text = node.toString();
```
hsivonen: It seems inefficient to create an intermediate `String` object compared to coping from a `char` array to a `char` array directly.

src/nu/validator/htmlparser/impl/TreeBuilder.java
```diff
@@ -426,6 +429,25 @@ public abstract class TreeBuilder<T> implements TokenHandler,
...
+
+    // Tracks if we've seen an option with the 'selected' attribute in the current select
+    private boolean seenSelectedOption = false;
+
+    // Tracks if we've already had an active option (first option was selected for cloning)
+    private boolean hadActiveOption = false;
+
+    // Stack to track the current parent in selectedcontent for element cloning
+    // When we're inside an active option, we push cloned elements here
+    private ArrayList<T> selectedContentCloneStack = new ArrayList<T>();
```
hsivonen: (After this lands, this bit will need tweaking to make the C++ translation work.)

hsivonen (changes requested):
Sorry about the delay.

As discussed on Matrix, I think the way this patch makes the cloning work on what the parser saw is incorrect and the spec instead requires the cloning to work on what's in the live DOM, which may be different if the `option` element opens, the network stalls, `setTimeout` modifies the DOM, and network resumes.

Further remarks in an inline comment.

src/nu/validator/htmlparser/impl/TreeBuilder.java
sideshowbarker: > @hsivonen wrote:
>
> Per spec, it seems that the cloning step runs on what's in the DOM and not on what the parser saw.
> 
> So I think the tree builder shouldn't keep track of `selectedcontent` but should have a hook for popping `option` to call a method that performs the cloning.
> 
> In Gecko, the method would queue a tree op to perform the cloning on the main-thread DOM.
> 
> The streaming SAX mode would ignore the cloning.
> 
> The other Java tree builders should override the method to perform the cloning on their respective tree models.

OK, https://github.com/validator/htmlparser/commit/e0ea675b9da5f54e197dad67bf8c43339e63925c attempts to correct the spec miss, and line up the rest of it all in the right way.

src/nu/validator/htmlparser/impl/TreeBuilder.java
```diff
@@ -2169,6 +2196,23 @@ public final void startTag(ElementName elementName,
                                 attributes = null; // CPP
                                 break starttagloop;
                             case HR:
+                                // Check if select is in scope
```
hsivonen: The spec seems to implicitly close "p" before the "select" check.

src/nu/validator/htmlparser/impl/TreeBuilder.java
```diff
@@ -2169,6 +2196,23 @@ public final void startTag(ElementName elementName,
...
+                                if (findLastInScope("select") != TreeBuilder.NOT_FOUND_ON_STACK) {
+                                    // Close any open option or optgroup first
+                                    if (isCurrent("option")) {
```
hsivonen: Why are "option" and "optgroup" popped separately instead of calling `generateImpliedEndTags();`?

src/nu/validator/htmlparser/impl/TreeBuilder.java
```diff
@@ -2184,7 +2228,27 @@ public final void startTag(ElementName elementName,
...
+                                        elementName, attributes,
+                                        formPointer);
+                                selfClosing = false;
+                                // [NOCPP[
+                                voidElement = true;
+                                // ]NOCPP]
+                                attributes = null; // CPP
+                                break starttagloop;
                             case INPUT:
+                                // Check if select is in scope and close it (for compatibility)
```
hsivonen: It seems that the fragment-case special case if context is "select" is missing.

src/nu/validator/htmlparser/impl/TreeBuilder.java
```diff
@@ -2184,7 +2228,27 @@ public final void startTag(ElementName elementName,
...
+                                eltPos = findLastInScope("select");
+                                if (eltPos != TreeBuilder.NOT_FOUND_ON_STACK) {
+                                    errStartTagWithSelectOpen(name);
+                                    while (currentPtr >= eltPos) {
+                                        pop();
+                                    }
+                                    resetTheInsertionMode();
```
hsivonen: Please add a comment that explains this.

src/nu/validator/htmlparser/impl/TreeBuilder.java
```diff
@@ -2196,6 +2260,16 @@ public final void startTag(ElementName elementName,
                                 attributes = null; // CPP
                                 break starttagloop;
                             case TEXTAREA:
+                                // Check if select is in scope and close it (for compatibility)
```
hsivonen: I don't find spec text corresponding to this. Why is this needed?

src/nu/validator/htmlparser/impl/TreeBuilder.java
```diff
@@ -2235,27 +2309,103 @@ public final void startTag(ElementName elementName,
                                 attributes = null; // CPP
                                 break starttagloop;
                             case SELECT:
+                                // Check if select is already in scope (nested select)
```
hsivonen: The fragment-case special case seems to be missing.

src/nu/validator/htmlparser/impl/TreeBuilder.java
```diff
@@ -426,6 +428,21 @@ public abstract class TreeBuilder<T> implements TokenHandler,
 
     private T headPointer;
 
+    // For customizable select: tracks the selectedcontent element to clone option content into
+    protected T selectedContentPointer;
```
hsivonen: Why do these still need to be tracked in the tree builder?

src/nu/validator/htmlparser/impl/TreeBuilder.java
```diff
@@ -2235,27 +2309,103 @@ public final void startTag(ElementName elementName,
...
+                                            elementName,
+                                            attributes, formPointer);
+                                    // No longer switch to IN_SELECT mode
+                                    attributes = null; // CPP
+                                    break starttagloop;
+                                }
+                            case OPTION:
+                                // Check if select is in scope
+                                if (findLastInScope("select") != TreeBuilder.NOT_FOUND_ON_STACK) {
+                                    // Reconstruct active formatting elements first
```
hsivonen: Why are the active formatting elements generated earlier than in the spec?

src/nu/validator/htmlparser/impl/TreeBuilder.java
```diff
@@ -2235,27 +2309,103 @@ public final void startTag(ElementName elementName,
...
+                                    reconstructTheActiveFormattingElements();
+                                    // Generate implied end tags except for optgroup
+                                    generateImpliedEndTagsExceptFor("optgroup");
+                                    // Check if option is in scope and close it
+                                    eltPos = findLastInScope("option");
+                                    if (eltPos != TreeBuilder.NOT_FOUND_ON_STACK) {
+                                        if (errorHandler != null && !isCurrent("option")) {
+                                            errUnclosedElementsImplied(eltPos, "option");
+                                        }
+                                        while (currentPtr >= eltPos) {
```
hsivonen: I don't see this in the spec. Why is this here?

src/nu/validator/htmlparser/impl/TreeBuilder.java
```diff
@@ -2235,27 +2309,103 @@ public final void startTag(ElementName elementName,
...
+                                if (findLastInScope("select") != TreeBuilder.NOT_FOUND_ON_STACK) {
+                                    // Generate implied end tags
+                                    generateImpliedEndTags();
+                                    // Check if optgroup is in scope and close it
+                                    eltPos = findLastInScope("optgroup");
+                                    if (eltPos != TreeBuilder.NOT_FOUND_ON_STACK) {
+                                        if (errorHandler != null && !isCurrent("optgroup")) {
+                                            errUnclosedElementsImplied(eltPos, "optgroup");
+                                        }
+                                        while (currentPtr >= eltPos) {
```
hsivonen: I don't see this in the spec. Why is this here?

src/nu/validator/htmlparser/impl/TreeBuilder.java
```diff
@@ -2235,27 +2309,103 @@ public final void startTag(ElementName elementName,
...
+                                // Check if select is in scope
+                                if (findLastInScope("select") != TreeBuilder.NOT_FOUND_ON_STACK) {
+                                    // Generate implied end tags
+                                    generateImpliedEndTags();
+                                    // Check if optgroup is in scope and close it
+                                    eltPos = findLastInScope("optgroup");
+                                    if (eltPos != TreeBuilder.NOT_FOUND_ON_STACK) {
+                                        if (errorHandler != null && !isCurrent("optgroup")) {
+                                            errUnclosedElementsImplied(eltPos, "optgroup");
+                                        }
```
hsivonen: It appears that also "option" being in scope is an error per spec.

src/nu/validator/htmlparser/impl/TreeBuilder.java
```diff
@@ -2329,11 +2479,25 @@ public final void startTag(ElementName elementName,
                                 attributes = null; // CPP
                                 break starttagloop;
                             case CAPTION:
-                            case COL:
-                            case COLGROUP:
                             case TBODY_OR_THEAD_OR_TFOOT:
                             case TR:
                             case TD_OR_TH:
+                                // Check if we're inside a select inside a table
```
hsivonen: How does this map to the spec?

src/nu/validator/htmlparser/impl/TreeBuilder.java
```diff
@@ -2346,6 +2510,26 @@ public final void startTag(ElementName elementName,
                                         attributes, formPointer);
                                 attributes = null; // CPP
                                 break starttagloop;
+                            case SELECTEDCONTENT:
```
hsivonen: It seems to me the tree builder shouldn't know about `selectedcontent` and looking for it is a DOM-side operation.

src/nu/validator/htmlparser/impl/TreeBuilder.java
```diff
@@ -3627,6 +3811,51 @@ public final void endTag(ElementName elementName) throws SAXException {
                                 }
                             }
                             break endtagloop;
+                        case OPTION:
```
hsivonen: It seems to me that "Any other end tag" should continue to handle these, except poping an "option" (regardless of tag) should trigger the DOM-side cloning.

src/nu/validator/htmlparser/impl/TreeBuilder.java
```diff
@@ -3674,6 +3903,22 @@ public final void endTag(ElementName elementName) throws SAXException {
                         case TEMPLATE:
                             // fall through to IN_HEAD;
                             break;
+                        case SELECT:
```
hsivonen: Why is this handled in a `select`-specific way instead of being bundled together with various block-ish elements?

src/nu/validator/htmlparser/impl/TreeBuilder.java
sideshowbarker: Fixed — `implicitlyCloseP()` now runs before the select-in-scope check, matching the spec order: _“If the stack of open elements has a `p` element in button scope, then close a `p` element”_ comes first.

src/nu/validator/htmlparser/impl/TreeBuilder.java
sideshowbarker: Fixed — now calls `generateImpliedEndTags()` per spec, then reports a parse error if `option` or `optgroup` is still in scope.

src/nu/validator/htmlparser/impl/TreeBuilder.java
sideshowbarker: In the relaxed-`select` model, the fragment case doesn't need a special guard for `input`: the context element isn't on the stack, so `findLastInScope("select")` returns `NOT_FOUND` and `input` is processed normally (inserted). This matches the html5lib tests that were updated for relaxed `select` (commit f994590 in html5lib-tests).

src/nu/validator/htmlparser/impl/TreeBuilder.java
sideshowbarker: Added spec-text comments throughout. The `resetTheInsertionMode()` call after popping `select` is needed because closing the `select` may leave the parser in a mode that no longer matches the stack (e.g., if `select` were inside a table cell, the mode needs to become `IN_CELL`).

src/nu/validator/htmlparser/impl/TreeBuilder.java
sideshowbarker: Mistake. Removed.

src/nu/validator/htmlparser/impl/TreeBuilder.java
sideshowbarker: Fixed. The fragment case is now checked first, before the select-in-scope check.

src/nu/validator/htmlparser/impl/TreeBuilder.java
sideshowbarker: They don't — removed. All `selectedcontent`/`option` tracking has been removed from `TreeBuilder`. Cloning is now handled entirely in the DOM-side subclasses via an `optionElementPopped()` hook that gets called from `pop()`.

src/nu/validator/htmlparser/impl/TreeBuilder.java
sideshowbarker: Fixed. `reconstructTheActiveFormattingElements()` now runs after the select-in-scope/else branch, matching the spec order.

src/nu/validator/htmlparser/impl/TreeBuilder.java
sideshowbarker: Mistake. Removed.

src/nu/validator/htmlparser/impl/TreeBuilder.java
sideshowbarker: Mistake. Removed.

src/nu/validator/htmlparser/impl/TreeBuilder.java
sideshowbarker: Fixed. The parse error now checks for both `option` and `optgroup` in scope, matching the spec: _“If the stack of open elements has an option element in scope or has an optgroup element in scope, then this is a parse error.”_

src/nu/validator/htmlparser/impl/TreeBuilder.java
sideshowbarker: Simplified. With the removal of `IN_SELECT_IN_TABLE`, these table-structure tags are just stray start tags in `IN_BODY`. They're now merged with the `COL`/`COLGROUP`/`FRAME`/`FRAMESET`/`HEAD` group as `errStrayStartTag` — which matches the spec.

src/nu/validator/htmlparser/impl/TreeBuilder.java
sideshowbarker: Agreed — the tree builder no longer knows about `selectedcontent`. The `optionElementPopped()` hook in `pop()` delegates to the DOM-side subclasses (`DOMTreeBuilder`, `XOMTreeBuilder`, `BrowserTreeBuilder`) where the ancestor walk and cloning live.

src/nu/validator/htmlparser/impl/TreeBuilder.java
sideshowbarker: Fixed. The dedicated `option`/`optgroup` end-tag cases have been removed. “Any other end tag” handles them. The `optionElementPopped()` hook in `pop()` triggers DOM-side cloning whenever an `option` element is popped — regardless of which tag caused the pop.

src/nu/validator/htmlparser/impl/TreeBuilder.java
sideshowbarker: Fixed. The `</select>` end tag is now folded into the block-element group (with `</div>`, `</fieldset>`, `</button>`, etc.) — matching the spec grouping.

src/nu/validator/htmlparser/impl/TreeBuilder.java
```diff
@@ -2170,6 +2156,33 @@ public final void startTag(ElementName elementName,
...
+                                            && (findLastInScope("option") != TreeBuilder.NOT_FOUND_ON_STACK
+                                                || findLastInScope("optgroup") != TreeBuilder.NOT_FOUND_ON_STACK)) {
+                                        errUnclosedElementsImplied(
+                                                findLastInScope("option") != TreeBuilder.NOT_FOUND_ON_STACK
+                                                        ? findLastInScope("option")
+                                                        : findLastInScope("optgroup"),
+                                                name);
+                                    }
+                                    // "3. Insert an HTML element for the token."
+                                    appendVoidElementToCurrent(elementName, attributes);
```
hsivonen: It looks like this is an extra insertion that shouldn't be here, since there's the pre-existing insertion after the closing brace.

src/nu/validator/htmlparser/impl/TreeBuilder.java
```diff
@@ -2235,31 +2273,97 @@ public final void startTag(ElementName elementName,
...
+                                //  element in scope:"
+                                eltPos = findLastInScope(name);
+                                if (eltPos != TreeBuilder.NOT_FOUND_ON_STACK) {
+                                    // "Parse error."
+                                    errStartSelectWhereEndSelectExpected();
+                                    // "Pop elements until a select element has been popped."
+                                    while (currentPtr >= eltPos) {
+                                        pop();
+                                    }
+                                    resetTheInsertionMode();
```
hsivonen: Why is `resetTheInsertionMode()` needed? I don't see it in the spec or in Keith's patch.

src/nu/validator/htmlparser/impl/TreeBuilder.java
```diff
@@ -2235,31 +2273,97 @@ public final void startTag(ElementName elementName,
...
+                                if (findLastInScope("select") != TreeBuilder.NOT_FOUND_ON_STACK) {
+                                    // "1. Generate implied end tags."
+                                    generateImpliedEndTags();
+                                    // "2. If the stack of open elements has an option element
+                                    //     in scope or has an optgroup element in scope, then
+                                    //     this is a parse error."
+                                    if (errorHandler != null
+                                            && (findLastInScope("option") != TreeBuilder.NOT_FOUND_ON_STACK
+                                                || findLastInScope("optgroup") != TreeBuilder.NOT_FOUND_ON_STACK)) {
+                                        if (findLastInScope("option") != TreeBuilder.NOT_FOUND_ON_STACK) {
```
hsivonen: It should  be possible to rearrange these checks so that the stack isn't walked twice. Perhaps makes sense as a follow-up to have less merging ahead.

src/nu/validator/htmlparser/impl/TreeBuilder.java
```diff
@@ -5116,13 +5046,19 @@ private void popForeign(int origPos, int eltPos) throws SAXException {
     private void silentPop() throws SAXException {
         StackNode<T> node = stack[currentPtr];
         assert debugOnlyClearLastStackSlot();
+        if (node.getGroup() == OPTION) {
```
hsivonen: I think it's impossible for this to happen, `since silentPop()` is only called in the `TEXT` insertion mode, where the element that caused the transition to `TEXT` is never `option`. Let's just assert here instead.

src/nu/validator/htmlparser/impl/TreeBuilder.java
```diff
@@ -5097,6 +5021,9 @@ private void popTemplateMode() {
     private void pop() throws SAXException {
         StackNode<T> node = stack[currentPtr];
         assert debugOnlyClearLastStackSlot();
+        if (node.getGroup() == OPTION) {
+            optionElementPopped(node.node);
```
hsivonen, line 5019: Considering the comment about `silentPop` it seems to me that it would be possible to handle `optionElementPopped` in `elementPopped` instead. That would mean handling in each subclass, though, so let's leave it this way.

gwt-src/nu/validator/htmlparser/gwt/BrowserTreeBuilder.java
```diff
@@ -474,4 +474,84 @@ private static native void removeChild(JavaScriptObject parent,
...
+                        return node.localName;
+                    }-*/;
+
+    private static native boolean hasAttribute(
+            JavaScriptObject node, String name) /*-{
+                        return node.hasAttribute(name);
+                    }-*/;
+
+    @Override
+    // https://html.spec.whatwg.org/multipage/form-elements.html#concept-select-option-selectedness
```
hsivonen: The spec link should be https://html.spec.whatwg.org/multipage/form-elements.html#maybe-clone-an-option-into-selectedcontent

gwt-src/nu/validator/htmlparser/gwt/BrowserTreeBuilder.java
```diff
@@ -474,4 +474,84 @@ private static native void removeChild(JavaScriptObject parent,
...
+                if ("select".equals(getLocalName(ancestor))) {
+                    select = ancestor;
+                    break;
+                }
+                ancestor = getParentNode(ancestor);
+            }
+            if (select == null) {
+                return;
+            }
+
```
hsivonen, line 522: It seems to me that there should be a `return` here if `select` has the attribute `multiple` set.

gwt-src/nu/validator/htmlparser/gwt/BrowserTreeBuilder.java
```diff
@@ -474,4 +474,84 @@ private static native void removeChild(JavaScriptObject parent,
...
+    private JavaScriptObject findDescendantByLocalName(
+            JavaScriptObject root, String localName) {
+        for (JavaScriptObject child = getFirstChild(root);
+                child != null; child = getNextSibling(child)) {
+            if (getNodeType(child) == 1) {
+                if (localName.equals(getLocalName(child))) {
+                    return child;
+                }
+                JavaScriptObject found = findDescendantByLocalName(
```
hsivonen: This would be better written iteratively than recursively. See https://github.com/validator/htmlparser/blob/69af110b44405e9777fd8d74efccd539119de630/src/nu/validator/saxtree/TreeParser.java#L85 for the iterative formulation.

src/nu/validator/htmlparser/dom/DOMTreeBuilder.java
```diff
@@ -354,4 +354,73 @@ protected Element createAndInsertFosterParentedElement(String ns, String name,
...
+                        select = elt;
+                        break;
+                    }
+                }
+                ancestor = ancestor.getParentNode();
+            }
+            if (select == null) {
+                return;
+            }
+
```
hsivonen, line 384: I think the `multiple` attribute needs to be checked here.

src/nu/validator/htmlparser/dom/DOMTreeBuilder.java
```diff
@@ -354,4 +354,73 @@ protected Element createAndInsertFosterParentedElement(String ns, String name,
...
+        for (Node child = root.getFirstChild(); child != null;
+                child = child.getNextSibling()) {
+            if (child.getNodeType() == Node.ELEMENT_NODE) {
+                Element elt = (Element) child;
+                if ("selectedcontent".equals(elt.getLocalName())
+                        && "http://www.w3.org/1999/xhtml".equals(
+                                elt.getNamespaceURI())) {
+                    return elt;
+                }
+                Element found = findSelectedContent(elt);
```
hsivonen: This, too, would be better as iterative than as recursive.

src/nu/validator/htmlparser/sax/SAXTreeBuilder.java
```diff
@@ -197,4 +198,111 @@ private Node previousSibling(Node table) {
...
+        if (select == null) {
+            return;
+        }
+
+        // Find the first <selectedcontent> descendant of <select>
+        Element selectedContent = findDescendant(select, "selectedcontent");
+        if (selectedContent == null) {
+            return;
+        }
+
```
hsivonen, line 220: Again `multiple`.

src/nu/validator/htmlparser/sax/SAXTreeBuilder.java
```diff
@@ -197,4 +198,111 @@ private Node previousSibling(Node table) {
...
+    private Element findDescendant(Element root, String localName) {
+        Node child = root.getFirstChild();
+        while (child != null) {
+            if (child.getNodeType() == NodeType.ELEMENT) {
+                Element elt = (Element) child;
+                if (localName.equals(elt.getLocalName())
+                        && "http://www.w3.org/1999/xhtml".equals(elt.getUri())) {
+                    return elt;
+                }
+                Element found = findDescendant(elt, localName);
```
hsivonen: Again iterative vs. recursive.

src/nu/validator/htmlparser/xom/XOMTreeBuilder.java
```diff
@@ -348,4 +348,70 @@ private int indexOfTable(Element table, Element stackParent) {
...
+                        select = elt;
+                        break;
+                    }
+                }
+                ancestor = ancestor.getParent();
+            }
+            if (select == null) {
+                return;
+            }
+
```
hsivonen, line 380: `multiple`

src/nu/validator/htmlparser/xom/XOMTreeBuilder.java
```diff
@@ -348,4 +348,70 @@ private int indexOfTable(Element table, Element stackParent) {
...
+        for (int i = 0; i < root.getChildCount(); i++) {
+            Node child = root.getChild(i);
+            if (child instanceof Element) {
+                Element elt = (Element) child;
+                if ("selectedcontent".equals(elt.getLocalName())
+                        && "http://www.w3.org/1999/xhtml".equals(
+                                elt.getNamespaceURI())) {
+                    return elt;
+                }
+                Element found = findSelectedContent(elt);
```
hsivonen: Iterative vs. recursive.

hsivonen (changes requested):
I'm sorry that I've been in a very bad time management loop.

This looks very good, but there are some minor issues.

Thank you!

gwt-src/nu/validator/htmlparser/gwt/BrowserTreeBuilder.java
sideshowbarker: > The spec link should be https://html.spec.whatwg.org/multipage/form-elements.html#maybe-clone-an-option-into-selectedcontent

Thanks — fixed (multiple places)

gwt-src/nu/validator/htmlparser/gwt/BrowserTreeBuilder.java
sideshowbarker: > It seems to me that there should be a `return` here if `select` has the attribute `multiple` set.

Yes — fixed

gwt-src/nu/validator/htmlparser/gwt/BrowserTreeBuilder.java
sideshowbarker: > @hsivonen wrote:
>
> This would be better written iteratively than recursively. See https://github.com/validator/htmlparser/blob/69af110b44405e9777fd8d74efccd539119de630/src/nu/validator/saxtree/TreeParser.java#L85 for the iterative formulation.

Very nice stuff — thanks. Made it so.

src/nu/validator/htmlparser/dom/DOMTreeBuilder.java
sideshowbarker: > This, too, would be better as iterative than as recursive.

Yup — now updated to that formulation

src/nu/validator/htmlparser/dom/DOMTreeBuilder.java
sideshowbarker: > I think the `multiple` attribute needs to be checked here.

yes, added

sideshowbarker:
@hsivonen Please ignore that review request. I pushed but I still have more changes to make in response to your review comments.

src/nu/validator/htmlparser/impl/TreeBuilder.java
sideshowbarker: > It looks like this is an extra insertion that shouldn't be here, since there's the pre-existing insertion after the closing brace.

yes, removed now

src/nu/validator/htmlparser/impl/TreeBuilder.java
sideshowbarker: > Why is `resetTheInsertionMode()` needed? I don't see it in the spec or in Keith's patch.

Not needed — it’s cruft. Removed.

src/nu/validator/htmlparser/impl/TreeBuilder.java
sideshowbarker: > It should  be possible to rearrange these checks so that the stack isn't
> walked twice. Perhaps makes sense as a follow-up to have less merging
> ahead.

Yes — went ahead and made the change.

src/nu/validator/htmlparser/impl/TreeBuilder.java
sideshowbarker: > I think it's impossible for this to happen, `since silentPop()` is only called in the `TEXT` insertion mode, where the element that caused the transition to `TEXT` is never `option`. Let's just assert here instead.

Made it so.

src/nu/validator/htmlparser/sax/SAXTreeBuilder.java
sideshowbarker: > Again iterative vs. recursive.

Yes, changed (in my earlier push)

src/nu/validator/htmlparser/sax/SAXTreeBuilder.java
sideshowbarker: > Again `multiple`.

Yes, changed

src/nu/validator/htmlparser/xom/XOMTreeBuilder.java
sideshowbarker: > Iterative vs. recursive.

Yes, changed

src/nu/validator/htmlparser/xom/XOMTreeBuilder.java
sideshowbarker: > @hsivonen wrote:
>
> `multiple`

Yes, changed.

sideshowbarker:
@hsivonen OK, all comments now resolved, I think. Ready for you to review again.

gwt-src/nu/validator/htmlparser/gwt/BrowserTreeBuilder.java
```diff
@@ -474,4 +474,98 @@ private static native void removeChild(JavaScriptObject parent,
...
+    // https://html.spec.whatwg.org/multipage/form-elements.html#maybe-clone-an-option-into-selectedcontent
+    // Implements "maybe clone an option into selectedcontent"
+    protected void optionElementPopped(JavaScriptObject option)
+            throws SAXException {
+        try {
+            // Find the nearest ancestor <select> element
+            JavaScriptObject ancestor = getParentNode(option);
+            JavaScriptObject select = null;
+            while (ancestor != null && getNodeType(ancestor) == 1) {
+                if ("select".equals(getLocalName(ancestor))) {
```
hsivonen: Unlike the other subclasses, this one does not appear to check the HTMLness of either `select` or `selectedcontent`. It's likely possible to fool at least the `selectedcontent` search with SVG. Can be a follow-up to fix.

hsivonen (approved):
Let's merge this. Thank you and sorry about delay.

gwt-src/nu/validator/htmlparser/gwt/BrowserTreeBuilder.java
sideshowbarker: I’ll take a shot at this before we merge

sideshowbarker: > Unlike the other subclasses, this one does not appear to check the HTMLness of either `select` or `selectedcontent`. It's likely possible to fool at least the `selectedcontent` search with SVG.

Fixed in https://github.com/validator/htmlparser/commit/39f24fcac8f5b4a71d0e8536c5e00464dd4c041c

hsivonen (approved):
LGTM. Thank you!

PR #118, opened 2026-05-12
Replace ICU4J UCharacter methods with standard Java Character class

sideshowbarker:
ICU4J 75.1 removed the char-taking overloads of UCharacter methods:

- isHighSurrogate(char)    — removed (only isHighSurrogate(int) remains)
- isLowSurrogate(char)     — removed (only isLowSurrogate(int) remains)
- getCodePoint(char, char) — removed (only getCodePoint(int, int) remains)

When code compiled against an older ICU4J (which had those methods) runs with ICU4J 75.1, it throws NoSuchMethodError — because the bytecode contains direct references to the removed method signatures.

Note: Code compiled against ICU4J 75.1 happens to work because Java’s implicit widening conversion (char → int) binds to the int versions at compile time. But that doesn’t help users of pre-compiled artifacts.

This change replaces all uses of those ICU4J methods with equivalent methods from Java’s standard Character class (available since Java 5):

- UCharacter.isHighSurrogate(c)   → Character.isHighSurrogate(c)
- UCharacter.isLowSurrogate(c)    → Character.isLowSurrogate(c)
- UCharacter.getCodePoint(c1, c2) → Character.toCodePoint(c1, c2)

### DIFF
--- a/src/nu/validator/htmlparser/extra/NormalizationChecker.java
+++ b/src/nu/validator/htmlparser/extra/NormalizationChecker.java
@@ -30,7 +30,6 @@ import org.xml.sax.Locator;
 import org.xml.sax.SAXException;
 import org.xml.sax.SAXParseException;
 
-import com.ibm.icu.lang.UCharacter;
 import com.ibm.icu.text.Normalizer;
 import com.ibm.icu.text.UnicodeSet;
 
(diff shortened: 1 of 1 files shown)


hsivonen (approved):
Thanks!

PR #119, opened 2026-05-13
ensure mode is always set to INITIAL during startTokenization

keithamus:
While currently, all branches set a mode, this is both subtle and fragile; if a branch ends up not setting mode it will never be reset. This means subsequent calls to start tokenisation can end up with a different initial mode.

This change simply resets mode to INITIAL on every startTokenization() call, ensuring that if we ever miss it, it will be reset.


hsivonen (approved):
Thanks!

PR #120, opened 2026-05-22
Ignore inputs inside a `select` fragment context.

keithamus:
This change does two things:

- updates the html5lib-tests, which exposes a failure in the current parser: it's missing the spec steps to ignore `<input>` when inside of a `<select>` during fragment parsing.
- subsequently fixes the failing test by adding the logic to ignore `<input>` inside of `<select>` during fragment parsing.

Refs https://github.com/html5lib/html5lib-tests/pull/195


hsivonen:
Thanks!

PR #121, opened 2026-05-25
Mozilla bug 2037935 - Special-case the most common element and attribute names in Speedometer3 innerHTML/DOMParser parsing.

hsivonen:
r? @smaug----

https://phabricator.services.mozilla.com/D302275


smaug---- (approved):
rs+

PR #122, opened 2026-05-25
Mozilla bug 2038124 - Add a flag to attribute names to indicate which attributes want atoms as value in Gecko.

hsivonen:
r? @smaug----

https://phabricator.services.mozilla.com/D302278


smaug---- (approved):
rs+, but if this gets still tweaked on Gecko side, perhaps some minor changes are needed? Probably not

PR #124, opened 2026-07-19
Initialize the document locator before starting tokenization

### LINKED ISSUE #123: The Document system identifier is null if you provide your own content handler.
I think the problem is that in `Driver.java`:

```
    public void tokenize(InputSource is, int bufferSize)
            throws SAXException, IOException {
        if (is == null) {
            throw new IllegalArgumentException("InputSource was null.");
        }
        tokenizer.start();
        confidence = Confidence.TENTATIVE;
        rewindableInputStream = null;
        tokenizer.initLocation(is.getPublicId(), is.getSystemId());
        ...
```

The call to `tokenizer.start()` initializes the HTML `Document` object with a null `Locator`. I think it would work correctly if `tokenizer.initLocation()` preceded `tokenizer.start()`.

sideshowbarker:
**Problem:** When a caller supplies its own `ContentHandler`, the Document system identifier is null.

**Cause:** `tokenizer.initLocation()` wasn’t getting called until after `Driver.tokenize()` called `tokenizer.start()`. So the tokenizer’s `systemId`/`publicId` were still null when `start()` delivered the first SAX events.

**Fix:** Call `initLocation()` before `start()`. The locator carries the input source’s public/system IDs before any downstream handler observes it. Fixes https://github.com/validator/htmlparser/issues/123.

### DIFF
--- a/src/nu/validator/htmlparser/io/Driver.java
+++ b/src/nu/validator/htmlparser/io/Driver.java
@@ -194,10 +194,10 @@ public class Driver implements EncodingDeclarationHandler {
         if (is == null) {
             throw new IllegalArgumentException("InputSource was null.");
         }
+        tokenizer.initLocation(is.getPublicId(), is.getSystemId());
         tokenizer.start();
         confidence = Confidence.TENTATIVE;
         rewindableInputStream = null;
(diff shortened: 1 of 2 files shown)


hsivonen (approved):
Thanks!

ndw:
Thank you!

PR #125, opened 2026-07-31
fix: Don’t report errors for omissible </option> end tag

sideshowbarker:
**Problem:** We unexpectedly reported errors for omitted `</option>` end tags in cases where the spec allows omitting them. See https://github.com/validator/validator/issues/2116.

**Cause:** The spec’s _“any other end tag”_ case for the _“in body”_ insertion mode requires generating implied end tags _“except for HTML elements with the same tag name as the token”_. But we didn’t implement that exception. Specifically, we called `generateImpliedEndTags()` if tag name is `"option"` but should instead call `generateImpliedEndTagsExceptFor("option")`.

**Fix:** Call `generateImpliedEndTagsExceptFor("option")` instead.


tenzap:
Please consider merging this since in Debian we depend on this htmlparser repo for the vnu validator package. And test now fails as long as this is not fixed in this htmlparser repo.

"file:tests/html/elements/option/omitted-end-tag-before-optgroup-isvalid.html":7:11: error: End tag “optgroup” seen, but there were open elements.

